PR #17, opened 2020-10-02
Refactoring capabilities

fiunchinho:
Towards: https://github.com/giantswarm/giantswarm/issues/13475

This PR adds 2 new validating webhooks for:

- node pool CREATE
- node pool UPGRADE

The webhooks check that the acceleratedNetworking flag is never changed during an upgrade and that is set to true on new node pools only if the instance type supports it.


config/config.go
```diff
@@ -35,13 +43,44 @@ func Parse() (Config, error) {
 		}
 	}
 
+	var resourceSkusClient compute.ResourceSkusClient
+	{
+		settings, err := auth.GetSettingsFromEnvironment()
+		if err != nil {
+			panic(err)
```
whites11, line 50: Can we maybe handle the error more gracefully?

config/config.go
```diff
@@ -35,13 +43,44 @@ func Parse() (Config, error) {
...
+		}
+		authorizer, err := settings.GetAuthorizer()
+		if err != nil {
+			panic(err)
```
whites11, line 54: Same here

whites11 (approved):
Apart from the panics, looks good to me

PR #24, opened 2020-10-05
Alpha upgrade block

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13598

This PR adds a few more objects to be validated in the azure validation webhook.
The goal is to avoid the possibility to upgrade from and to any release that contains the word `alpha` in it's name.


whites11:
```
$ kubectl --context=giantswarm-godsmack -n org-giantswarm edit cluster f7ui4
error: clusters.cluster.x-k8s.io "f7ui4" could not be patched: admission webhook "cluster.update.azure-admission-controller-unique.giantswarm.io" denied the request: invalid operation error: It is not possible to upgrade to or from an alpha release
You can run `kubectl replace -f /tmp/kubectl-edit-xi8o9.yaml` to try this update again.

$ kubectl --context=giantswarm-godsmack -n org-giantswarm edit azurecluster f7ui4
error: azureclusters.infrastructure.cluster.x-k8s.io "f7ui4" could not be patched: admission webhook "azureclusters.update.azure-admission-controller-unique.giantswarm.io" denied the request: invalid operation error: It is not possible to upgrade to or from an alpha release
You can run `kubectl replace -f /tmp/kubectl-edit-fho1t.yaml` to try this update again.

$ kubectl --context=giantswarm-godsmack -n org-giantswarm edit azuremachine f7ui4-master-0 
error: azuremachines.infrastructure.cluster.x-k8s.io "f7ui4-master-0" could not be patched: admission webhook "azuremachines.update.azure-admission-controller-unique.giantswarm.io" denied the request: invalid operation error: It is not possible to upgrade to or from an alpha release
You can run `kubectl replace -f /tmp/kubectl-edit-s4d2m.yaml` to try this update again.

```

pkg/azureupdate/common.go
```diff
@@ -68,6 +69,11 @@ func upgradeAllowed(ctx context.Context, g8sclient versioned.Interface, oldVersi
 			return false, microerror.Maskf(invalidOperationError, "downgrading is not allowed (attempted to downgrade from %s to %s)", oldVersion, newVersion)
 		}
 
+		// Check if either version is an alpha one.
+		if isAlphaRelease(oldVersion.String()) || isAlphaRelease(newVersion.String()) {
```
tuommaki, line 73: Would it make any sense to block the upgrade if release name isn't plain semver `XX.YY.ZZ`? For now we don't have RCs or anything similar where this upgrade would be useful.

whites11: Not sure about this.
I can imagine a future where we do beta releases that have upgrades allowed.
Plus we sometimes use named releases such as `13.0.0-yulia` and we might want to test upgrades from and to those releases.

pkg/azurecluster/update.go
```diff
@@ -0,0 +1,91 @@
...
+	"k8s.io/api/admission/v1beta1"
+	restclient "k8s.io/client-go/rest"
+	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
+
+	"github.com/giantswarm/azure-admission-controller/pkg/validator"
+)
+
+type UpdateValidator struct {
+	k8sClient k8sclient.Interface
```
fiunchinho, line 23: There is a `k8sClient` defined on all `UpdateValidator`.  I don't think it's needed and it could be removed.

pkg/azuremachine/update.go
```diff
@@ -0,0 +1,91 @@
...
+	AzureMachineOldCR := &capzv1alpha3.AzureMachine{}
+	if _, _, err := validator.Deserializer.Decode(request.Object.Raw, nil, AzureMachineNewCR); err != nil {
+		return false, microerror.Maskf(parsingFailedError, "unable to parse AzureMachine CR: %v", err)
+	}
+	if _, _, err := validator.Deserializer.Decode(request.OldObject.Raw, nil, AzureMachineOldCR); err != nil {
+		return false, microerror.Maskf(parsingFailedError, "unable to parse AzureMachine CR: %v", err)
+	}
+
+	oldClusterVersion := AzureMachineOldCR.Labels[label.ReleaseVersion]
+	newClusterVersion := AzureMachineNewCR.Labels[label.ReleaseVersion]
```
fiunchinho, line 74: Should we consider the case when the label is not present?

whites11: uhm yeah maybe. How should we behave in such case?

pkg/azurecluster/update.go
whites11: removed

pkg/azuremachine/update.go
whites11: OK I have an idea

main.go
```diff
@@ -94,51 +97,99 @@ func mainError() error {
...
 	}
 
-	azureConfigValidatorConfig := azureupdate.AzureConfigValidatorConfig{
-		K8sClient: k8sClient,
-		Logger:    newLogger,
-	}
-	azureConfigValidator, err := azureupdate.NewAzureConfigValidator(azureConfigValidatorConfig)
-	if err != nil {
-		return microerror.Mask(err)
+	var azureConfigValidator *azureupdate.AzureConfigValidator
```
tuommaki, line 100: I'm totally ok with this if you want to have separate validators for each CRD, but for this specific upgrade use case on releases, you should be able to write a single generic validator for all CRs. Instead of deserializing into specific CR, maybe you could deserialize into [ObjectMeta](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta)? From there you can do `m.GetLabels()` and have a unified implementation.

There might be cases in the future where having distinct validators for specific types helps with smaller PRs, but I leave the judgment to you. Either way works I guess.

pkg/azurecluster/update.go
```diff
@@ -0,0 +1,60 @@
...
+	azureClusterNewCR := &capzv1alpha3.AzureCluster{}
+	azureClusterOldCR := &capzv1alpha3.AzureCluster{}
+	if _, _, err := validator.Deserializer.Decode(request.Object.Raw, nil, azureClusterNewCR); err != nil {
+		return false, microerror.Maskf(parsingFailedError, "unable to parse AzureCluster CR: %v", err)
+	}
+	if _, _, err := validator.Deserializer.Decode(request.OldObject.Raw, nil, azureClusterOldCR); err != nil {
+		return false, microerror.Maskf(parsingFailedError, "unable to parse AzureCluster CR: %v", err)
+	}
+
+	oldClusterVersion := azureClusterOldCR.Labels[label.ReleaseVersion]
```
tuommaki, line 42: The specific logic of validating release labels of two objects should probably be extracted into own package and used in all validators. That way it's easier to change later if we want (and we might want). WDYT? It could accept [metav1.Object](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Object) interface as param and do the thing from there.

pkg/azureupdate/common.go
tuommaki: Ack. I'm fine with that.

tuommaki (approved):
Overall I feel this is fine. Not blocking here, but providing few notes in case you see them as improvements.

pkg/azurecluster/update.go
tuommaki: Oh, yeah, another thought: This way you could easily write test cases without too much boilerplate or setup. :slightly_smiling_face:

whites11: Yes I am trying to remove some code duplication with a similar approach. Let's see how it turns out

whites11:
@fiunchinho @tuommaki I refactored the whole thing in order to have a common package implementing the upgrade check logic.
This leads to a bigger diff in the PR but should make it a better result in the end.

internal/releaseversion/releaseversion.go
```diff
@@ -0,0 +1,104 @@
...
+func GetVersionFromString(version string) (semver.Version, error) {
+	return semver.ParseTolerant(version)
+}
+
+func IsAlphaRelease(release string) bool {
+	return strings.Contains(release, "alpha")
+}
+
+func UpgradeAllowed(ctx context.Context, g8sclient versioned.Interface, oldVersion semver.Version, newVersion semver.Version) (bool, error) {
+	if !oldVersion.Equals(newVersion) {
```
fiunchinho, line 37: small style comment: I believe it's easier to read returning earlier and one less indentation level
```
if oldVersion.Equals(newVersion) {
    return true, nil
}

availableReleases, err := availableReleases(ctx, g8sclient)
...
..
.
```

main.go
```diff
@@ -94,51 +97,102 @@ func mainError() error {
...
+	{
+		azureConfigValidatorConfig := azureupdate.AzureConfigValidatorConfig{
+			K8sClient: k8sClient,
```
fiunchinho, line 103: We've been moving towards using the controller-runtime client everywhere. Wherever we can use it, I'd say let's use it. The different `Config` objects would depend on client.Client and you would just need to pass `k8sclient.CtrlClient()` instead of `k8sclient`.

pkg/azurecluster/update.go
```diff
@@ -0,0 +1,60 @@
...
+	azureClusterNewCR := &capzv1alpha3.AzureCluster{}
+	azureClusterOldCR := &capzv1alpha3.AzureCluster{}
+	if _, _, err := validator.Deserializer.Decode(request.Object.Raw, nil, azureClusterNewCR); err != nil {
+		return false, microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureCluster CR: %v", err)
+	}
+	if _, _, err := validator.Deserializer.Decode(request.OldObject.Raw, nil, azureClusterOldCR); err != nil {
+		return false, microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureCluster CR: %v", err)
+	}
+
+	oldClusterVersion, err := releaseversion.GetVersionFromCRLabels(azureClusterOldCR.Labels)
```
fiunchinho, line 46: We are calling `GetVersionFromCRLabels` (twice) and `UpgradeAllowed` from the `releaseversion` package. I was thinking if it would be better to add a new function `releaseversion.Validate()` and just call that? All the validations related to releases/versioning would be done there. If we were to add validations based on other things they would be in other packages. And the `UpdateValidator` would look something like

```Validate() {
    err = releaseversion.Validate()
    if err != nil {
        return false, microerror.mask(err)
    }

    err = anotherpackage.Validate()
    if err != nil {
        return false, microerror.mask(err)
    }
}```

File: main.go (hunk 4 above)
Comment (whites11) [reply main.go:100]: I like the idea of abstracting the specific CR when only the ObjectMeta is needed.
But it's likely that in the near future we'll need to deal with the whole CR for validation purposes, so I guess it's ok to work with the full objects already.

File: internal/releaseversion/releaseversion.go (hunk 6 above)
Comment (whites11) [reply internal/releaseversion/releaseversion.go]: Agreed, done

File: pkg/azurecluster/update.go (hunk 8 above)
Comment (whites11) [reply pkg/azurecluster/update.go:46]: ok that makes sense

Comment (whites11) [reply pkg/azurecluster/update.go]: I extracted the parsing of release to another shared module and renamed the validating function to Validate.
I don't like the idea of passing the CRs to the Validate function because it would mean it should know how to extract the release information from all CR types.
WDYT?

File: main.go (hunk 7 above)
Comment (whites11) [reply main.go:103]: but how do I get a G8sClient out of a CtrlClient?

Comment (tuommaki) [reply main.go:103]: You don't need `G8sClient` anymore. CtrlClient works by looking into the object that is given to it (which it will then fill w/ data).

e.g. AzureConfigs:
```go
azureConfig := &providerv1alpha1.AzureConfig{}
err := ctrlClient.Get(ctx, client.ObjectKey{Name: "mycluster", Namespace: "default"}, azureConfig)
if err != nil {
    panic("ftw")
}
```

whites11: ok thanks let me try

pkg/azurecluster/update.go
fiunchinho: Yeah, I see what you mean. `AzureConfig` and `AzureClusterConfig` are not labeled with the version right? If they are not, it's fine like you have it now.

main.go
whites11: switched to ctrlClient everywhere

whites11:
Merging

PR #39, opened 2020-10-13
Block changing node pool instance type from one that supports premium…

whites11:
Towards: giantswarm/giantswarm#13699

This PR avoids changing the instance type for a node pool from one supporting premium storage to one that doesn't support it.

This simply can't work.


tuommaki (commented):
Related to `azure-operator` PR, would it make sense to also guard for that `OSDisk` storage type change in this PR?

whites11:
> Related to `azure-operator` PR, would it make sense to also guard for that `OSDisk` storage type change in this PR?

I guess you are referring to the `StorageAccountType` fields in the CR.
Those fields are not utilized in the current implementation.
The azure operator checks for the right storage account type considering VM type support and completely ignores the CR fields.
It could make sense to use the field and remove all logic from the azure operator (that will trust and use the CR field) and leave all the checks in the admission controller.

WDYAT?

pkg/azuremachinepool/validate_azuremachinepool_update.go
```diff
@@ -53,6 +53,24 @@ func (a *UpdateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
...
+		if err != nil {
+			return false, microerror.Mask(err)
+		}
+
+		if oldPremium && !newPremium {
+			// We can't downgrade from a VM type supporting premium storage to one that doesn't.
+			// Azure doesn't support that.
+			return false, microerror.Maskf(invalidOperationError, "Changing the node pool VM type from one that supports accelerated networking to one that does not is unsupported.")
+		}
+	}
```
fiunchinho, line 72: Could we move this code to a function called `checkPremiumStorageUpdateIsValid` or something similar and call it from here? I think the `Validate` function reads better if we have separate calls for different things we want to check.

pkg/mutator/handler.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+var (
+	scheme        = runtime.NewScheme()
+	codecs        = serializer.NewCodecFactory(scheme)
+	Deserializer  = codecs.UniversalDeserializer()
+	InternalError = errors.New("internal admission controller error")
+)
+
+func Handler(mutator Mutator) http.HandlerFunc {
```
fiunchinho, line 33: I'm surprised we have to write this code ourselves. I expected kubernetes to provide some skaffolding for this.

pkg/azuremachinepool/validate_azuremachinepool_update.go
whites11: Done

pkg/azuremachinepool/storageaccounttype.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/giantswarm/azure-admission-controller/internal/vmcapabilities"
+)
+
+func checkStorageAccountTypeIsValid(ctx context.Context, vmcaps *vmcapabilities.VMSKU, azureMachinePool *expcapzv1alpha3.AzureMachinePool) error {
+	selectedStorageAccount := azureMachinePool.Spec.Template.OSDisk.ManagedDisk.StorageAccountType
+
+	if selectedStorageAccount != string(compute.StorageAccountTypesStandardLRS) &&
+		selectedStorageAccount != string(compute.StorageAccountTypesPremiumLRS) {
+		// Storage account type is invalid.
+		return microerror.Maskf(invalidOperationError, "Storage account type '%s' is invalid. Allowed values are %s and %s", selectedStorageAccount, string(compute.StorageAccountTypesStandardLRS), string(compute.StorageAccountTypesPremiumLRS))
```
tuommaki, line 19:
```suggestion
		return microerror.Maskf(invalidOperationError, "Storage account type %q is invalid. Allowed values are %q and %q", selectedStorageAccount, string(compute.StorageAccountTypesStandardLRS), string(compute.StorageAccountTypesPremiumLRS))
```

> %q	a double-quoted string safely escaped with Go syntax

That way it should read something like following:
`Storage account type "foobarLRS" is invalid. Allowed values are "StandardLRS" and "PremiumLRS"`

pkg/mutator/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	return func(writer http.ResponseWriter, request *http.Request) {
+		if request.Header.Get("Content-Type") != "application/json" {
+			mutator.Log("level", "error", "message", fmt.Sprintf("invalid content-type: %s", request.Header.Get("Content-Type")))
```
tuommaki, line 36:
```suggestion
			mutator.Log("level", "error", "message", fmt.Sprintf("invalid content-type: %q", request.Header.Get("Content-Type")))
```

pkg/mutator/handler.go
```diff
@@ -0,0 +1,125 @@
...
+
+		patch, err := mutator.Mutate(request.Context(), review.Request)
+		if err != nil {
+			writeResponse(mutator, writer, errorResponse(review.Request.UID, microerror.Mask(err)))
+			return
+		}
+
+		patchData, err := json.Marshal(patch)
+		if err != nil {
+			mutator.Log("level", "error", "message", fmt.Sprintf("unable to serialize patch for %s: %v", resourceName, err))
```
tuommaki, line 64:
```suggestion
			mutator.Log("level", "error", "message", fmt.Sprintf("unable to serialize patch for %s", resourceName), "stack", microerror.JSON(err))
```

pkg/mutator/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	return func(writer http.ResponseWriter, request *http.Request) {
+		if request.Header.Get("Content-Type") != "application/json" {
+			mutator.Log("level", "error", "message", fmt.Sprintf("invalid content-type: %s", request.Header.Get("Content-Type")))
+			writer.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		data, err := ioutil.ReadAll(request.Body)
+		if err != nil {
+			mutator.Log("level", "error", "message", "unable to read request")
```
tuommaki, line 43:
```suggestion
			mutator.Log("level", "error", "message", "unable to read request", "stack", microerror.JSON(err))
```

pkg/mutator/handler.go
```diff
@@ -0,0 +1,125 @@
...
+			writer.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		review := v1beta1.AdmissionReview{}
+		if _, _, err := Deserializer.Decode(data, nil, &review); err != nil {
+			mutator.Log("level", "error", "message", "unable to parse admission review request")
```
tuommaki, line 50:
```suggestion
			mutator.Log("level", "error", "message", "unable to parse admission review request", "stack", microerror.JSON(err))
```

pkg/mutator/handler.go
```diff
@@ -0,0 +1,125 @@
...
+func writeResponse(mutator Mutator, writer http.ResponseWriter, response *v1beta1.AdmissionResponse) {
+	resp, err := json.Marshal(v1beta1.AdmissionReview{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AdmissionReview",
+			APIVersion: "admission.k8s.io/v1",
+		},
+		Response: response,
+	})
+	if err != nil {
+		mutator.Log("level", "error", "message", "unable to serialize response", microerror.JSON(err))
```
tuommaki, line 109:
```suggestion
		mutator.Log("level", "error", "message", "unable to serialize response", "stack", microerror.JSON(err))
```

pkg/mutator/handler.go
```diff
@@ -0,0 +1,125 @@
...
+		writer.WriteHeader(http.StatusInternalServerError)
+	}
+	if _, err := writer.Write(resp); err != nil {
+		mutator.Log("level", "error", "message", "unable to write response", microerror.JSON(err))
```
tuommaki, line 113:
```suggestion
		mutator.Log("level", "error", "message", "unable to write response", "stack", microerror.JSON(err))
```

pkg/validator/handler.go
```diff
@@ -77,6 +77,8 @@ func writeResponse(validator Validator, writer http.ResponseWriter, response *v1
 	if _, err := writer.Write(resp); err != nil {
 		validator.Log("level", "error", "message", "unable to write response", microerror.JSON(err))
 	}
+
+	validator.Log("level", "info", "message", fmt.Sprintf("Validating request responded with result: %t", response.Allowed))
```
tuommaki, line 81:
```suggestion
	validator.Log("level", "info", "message", fmt.Sprintf("validated request responded with result: %t", response.Allowed))
```
Isn't the validation already done at this point when there's result? Hence past-tence :slightly_smiling_face: 

I was also wondering if this is useful `info` level log or rather a `debug` one?

Also: Above line 78 seems to miss `"stack"` key before error value.

pkg/azuremachinepool/mutate_azuremachinepool_create.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return m, nil
+}
+
+func (m *CreateMutator) Mutate(ctx context.Context, request *v1beta1.AdmissionRequest) ([]mutator.PatchOperation, error) {
+	var result []mutator.PatchOperation
+
+	if request.DryRun != nil && *request.DryRun {
+		return result, nil
```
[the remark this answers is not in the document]
tuommaki: Potentially noting something like: _Dry run is not supported. Request processing stopped._

tuommaki (commented):
Code looks mostly fine. Few suggestions around logging.

pkg/validator/handler.go
whites11: good catch, thank you

fiunchinho (approved):
LGTM

PR #40, opened 2020-10-15
Avoid ssh keys

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13783

This PR check the ssh key fields in azuremachine and azuremachinepool are empty.
We don't use em and don't want the customer to think they can


pkg/azuremachine/validate_create.go
```diff
@@ -0,0 +1,58 @@
...
+	"k8s.io/api/admission/v1beta1"
+	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-admission-controller/internal/errors"
+	"github.com/giantswarm/azure-admission-controller/pkg/validator"
+)
+
+type CreateValidator struct {
+	ctrlClient client.Client
```
fiunchinho, line 17: I don't think you are using the client?

pkg/azuremachine/validate_update.go
```diff
@@ -51,6 +51,11 @@ func (a *UpdateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
 		return false, microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureMachine CR: %v", err)
 	}
 
+	err := checkSSHKeyIsEmpty(ctx, AzureMachineNewCR)
```
fiunchinho, line 54: I know it's not from this PR but `AzureMachineNewCR` should start with lowercase.

fiunchinho (commented):
Would it be possible to add some simple tests?

pkg/azuremachine/validate_create.go
whites11: removed

pkg/azuremachine/validate_update.go
whites11: renamed

whites11:
> Would it be possible to add some simple tests?

done

fiunchinho (approved):
you rock

PR #42, opened 2020-10-16
Block configuring AzureMachinePool data disks, and set default value instead

axbarsan:
Towards giantswarm/giantswarm#13822

- Set default values in the `AzureMachinePool.DataDisks` property on resource creation
- Validate that the `AzureMachinePool.DataDisks` property has the right value during resource **creation**
- Validate that the `AzureMachinePool.DataDisks` property has the right value during resource **update**


pkg/azuremachinepool/datadisks.go
```diff
@@ -0,0 +1,32 @@
...
+	{
+		NameSuffix: "kubelet",
+		DiskSizeGB: 100,
+		Lun:        to.Int32Ptr(22),
+	},
+}
+
+func checkDataDisks(ctx context.Context, mp *expcapzv1alpha3.AzureMachinePool) error {
+	if !reflect.DeepEqual(mp.Spec.Template.DataDisks, desiredDataDisks) {
+		return microerror.Maskf(invalidOperationError, "AzureMachinePool.Spec.Template.DataDisks does not have required value.")
```
whites11, line 28:
```suggestion
		return microerror.Maskf(invalidOperationError, "AzureMachinePool.Spec.Template.DataDisks is reserved and should not be set.")
```

whites11: This is the message the customer will read if they set the field, better be explicit why they can't set the value for it.

whites11 (approved):
Apart from the error message that could be more explicit (left a suggestion) LGTM.
Well done!

pkg/azuremachinepool/validate_azuremachinepool_update_test.go
```diff
@@ -41,92 +41,110 @@ func TestAzureMachinePoolUpdateValidate(t *testing.T) {
 	testCases := []testCase{
 		{
 			name:         "case 0: AcceleratedNetworking is enabled in CR and we don't change it or the instance type",
-			oldNodePool:  azureMPRawObject(supportedInstanceType[0], &tr, string(compute.StorageAccountTypesStandardLRS)),
-			newNodePool:  azureMPRawObject(supportedInstanceType[0], &tr, string(compute.StorageAccountTypesStandardLRS)),
+			oldNodePool:  azureMPRawObject(supportedInstanceType[0], &tr, string(compute.StorageAccountTypesStandardLRS), nil),
```
fiunchinho, line 44: What do you think about passing `desiredDatadisks` instead of `nil` and removing the `nil` check from the function? I feel that always passing the disks is more explicit and we don't have to know that `nil` means `desiredDatadisks`.

fiunchinho (approved):

PR #49, opened 2020-10-22
Ensure that organization label value is normalized in Azure CRs

tuommaki:
This change ensures that organization label value is normalized when new CR
objects are created for Azure clusters.


tuommaki:
Tested on `ghost` with an `AzureMachine` as well and this seems to work now.

fiunchinho (changes requested):
Update the changelog please! otherwise LGTM

axbarsan (approved):

fiunchinho (approved):

PR #50, opened 2020-10-22
Base domain

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13916

This PR ensure a default value is provided for `AzureCluster.ControlPlaneEndpoint`, `Cluster.ControlPlaneEndpoint` and `Cluster.ClusterNetwork fields`.

It also ensures those fields do not change when the CRs are updated.

Finally, it ensures that any values provided by the customer at creation time are the expected values.

This will allow us to get rid of the `--domain` flag in `kubectl-gs`


pkg/azurecluster/validate_create.go
```diff
@@ -0,0 +1,56 @@
...
+func NewCreateValidator(config CreateValidatorConfig) (*CreateValidator, error) {
+	if config.BaseDomain == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.BaseDomain must not be empty", config)
+	}
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+
+	v := &CreateValidator{
+		logger: config.Logger,
```
fiunchinho, line 34: You are not setting the `baseDomain`

pkg/cluster/control_plane_endpoint.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if cluster.Spec.ClusterNetwork.ServiceDomain != key.ServiceDomain(cluster.Name, baseDomain) {
+		return microerror.Maskf(errors.InvalidOperationError, "ClusterNetwork.ServiceDomain can only be set to %s", key.ServiceDomain(cluster.Name, baseDomain))
+	}
+
+	if cluster.Spec.ClusterNetwork.Services == nil {
+		return microerror.Maskf(errors.InvalidOperationError, "ClusterNetwork.Services can't be null")
+	}
+
+	if !reflect.DeepEqual(cluster.Spec.ClusterNetwork.Services.CIDRBlocks, []string{key.ClusterNetworkServiceCIDR}) {
```
fiunchinho, line 30: Will be the length always 1? I'm asking because order matters, but I believe it's fine because it will be only one.

pkg/cluster/control_plane_endpoint.go
```diff
@@ -0,0 +1,56 @@
...
+		return microerror.Maskf(errors.InvalidOperationError, "ClusterNetwork.Services.CIDRBlocks can only be set to [%s]", key.ClusterNetworkServiceCIDR)
+	}
+
+	return nil
+}
+
+func validateControlPlaneEndpoint(cluster capiv1alpha3.Cluster, baseDomain string) error {
+	host := key.GetControlPlaneEndpointHost(cluster.Name, baseDomain)
+	if cluster.Spec.ControlPlaneEndpoint.Host != host {
+		return microerror.Maskf(errors.InvalidOperationError, "ControlPlaneEndpoint.Port can only be set to %s", host)
```
fiunchinho, line 40:
```suggestion
		return microerror.Maskf(errors.InvalidOperationError, "ControlPlaneEndpoint.Host can only be set to %s", host)
```

pkg/cluster/mutate_create.go
```diff
@@ -0,0 +1,125 @@
...
+
+		return mutator.PatchAdd("/spec/clusterNetwork", clusterNetwork), nil
+	}
+
+	return nil, nil
+}
+
+func (m *CreateMutator) ensureControlPlaneEndpointHost(ctx context.Context, clusterCR *capiv1alpha3.Cluster) (*mutator.PatchOperation, error) {
+	if clusterCR.Spec.ControlPlaneEndpoint.Host == "" {
+		return mutator.PatchAdd("/spec/controlPlaneEndpoint/host", fmt.Sprintf("api.%s.%s", clusterCR.Name, m.baseDomain)), nil
```
fiunchinho, line 113: should `fmt.Sprintf("api.%s.%s", clusterCR.Name, m.baseDomain)` be returned by a function in the `key` package like the rest of the values? WDYT?

pkg/cluster/mutate_create.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return nil, nil
+}
+
+func (m *CreateMutator) ensureControlPlaneEndpointPort(ctx context.Context, clusterCR *capiv1alpha3.Cluster) (*mutator.PatchOperation, error) {
+	if clusterCR.Spec.ControlPlaneEndpoint.Port == 0 {
+		return mutator.PatchAdd("/spec/controlPlaneEndpoint/port", 443), nil
```
fiunchinho, line 121: Same as above.

pkg/cluster/mutate_create_test.go
```diff
@@ -0,0 +1,165 @@
...
+			},
+			errorMatcher: nil,
+		},
+		{
+			name:         fmt.Sprintf("case 1: ControlPlaneEndpoint has a value"),
+			cluster:      clusterRawObject("ab132", clusterNetwork, "api.giantswarm.io", 123),
+			patches:      []mutator.PatchOperation{},
+			errorMatcher: nil,
+		},
+		// This test doesn't work because the clusterNetwork struct has pointers in it and we can't compare them.
```
fiunchinho, line 63: So let's remove it? :smile:

pkg/cluster/validate_create.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	if config.CtrlClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.CtrlClient must not be empty", config)
+	}
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+
+	v := &CreateValidator{
+		ctrlClient: config.CtrlClient,
```
fiunchinho, line 40: You are not setting the `baseDomain`. And it seems the client is not used.

fiunchinho (commented):
I left some small comments. Looking good!

pkg/azurecluster/validate_create.go
whites11: good catch, thanks a lot

pkg/cluster/control_plane_endpoint.go
whites11: it is for now

pkg/cluster/mutate_create.go
whites11: yeah totally

pkg/cluster/mutate_create_test.go
whites11: having that test would be useful, but we can't implement it.
I'd like to keep track of the need of having it though.
Is this an hard blocker for you?

pkg/cluster/validate_create.go
whites11: :( thanks

pkg/cluster/mutate_create.go
whites11: done

pkg/cluster/mutate_create_test.go
fiunchinho: not at all, let's keep it then

fiunchinho (approved):

PR #51, opened 2020-10-23
Prevent Organization label change on UPDATE

tuommaki:
Once the cluster has been created in given organization, it can't be changed
anymore. Therefore it makes sense to prevent any updates to its value as well.


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/giantswarm/k8sclient/v4 v4.0.0
 	github.com/giantswarm/microerror v0.2.1
 	github.com/giantswarm/micrologger v0.3.3
+	github.com/giantswarm/to v0.3.0
```
fiunchinho, line 17: We are already depending on `autorest/to`. Let's use that instead of adding another dependency.

fiunchinho (changes requested):
Your friend the Changelog needs an update :jack_o_lantern:

tuommaki:
> Your friend the Changelog needs an update

Oops!

tuommaki:
After the organization name normalization was removed. Do we need this? This is still valid restriction, but I'm not sure if this is strictly necessary?

whites11 (approved):
LGTM

PR #56, opened 2020-10-26
Don't normalize organization label value

tuommaki:
Due to legacy reasons, the use of case-sensitive organization label value has
spread quite widely and it seems to be a better approach to not normalize it in
order to maintain existing behavior on other system components.


whites11 (approved):
Seems legit

fiunchinho (approved):

nprokopic (approved):

PR #58, opened 2020-10-27
Validate cluster conditions

nprokopic:
Towards https://github.com/giantswarm/giantswarm/issues/14048

`Cluster` `Creating` validations:
- setting condition status to `Unknown` is not allowed
- newly set condition status must be either `True` or `False`
- removing condition that was already set is not allowed
- changing condition status from `False`to `True` is not allowed

`Cluster` `Upgrading` validations:
- setting condition status to `Unknown` is not allowed
- newly set condition status must be either `True` or `False`
- removing condition that was already set is not allowed

Other:
- Removed legacy `AzureConfig` conditions check when upgrading clusters
- Added `Cluster` conditions check when upgrading cluster, so that cluster upgrade cannot be initiated while the cluster is being created or upgraded


internal/conditions/valdiatecluster.go
```diff
@@ -0,0 +1,122 @@
...
+
+func ValidateUpgradingCondition(oldClusterCR *capi.Cluster, newClusterCR *capi.Cluster) error {
+	newUpgradingCondition := capiconditions.Get(newClusterCR, aeconditions.UpgradingCondition)
+
+	// Setting Unknown is not allowed, we should be able to tell if the cluster
+	// is being upgraded or not.
+	if newUpgradingCondition != nil && newUpgradingCondition.Status == corev1.ConditionUnknown {
+		errorMessage := "Setting Unknown status to Upgrading conditions is not allowed"
+		return microerror.Maskf(errors.InvalidConditionModificationError, errorMessage)
+	}
```
nprokopic, line 52: - `newUpgradingCondition != nil`: `Upgrading` condition is set to some value
- `newUpgradingCondition.Status == corev1.ConditionUnknown`: and it is set to a condition with `Unknown` status

So we block this, as we don't allow setting on `Unknown` condition, which can be in theory used when we don't know if it should be `True` or `False` (and we should always know).

internal/conditions/valdiatecluster.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// We only allow True and False as condition status values. Any other new
+	// or custom condition status values are not allowed.
+	conditionStatusNotAllowed := newUpgradingCondition != nil &&
+		newUpgradingCondition.Status != corev1.ConditionTrue &&
+		newUpgradingCondition.Status != corev1.ConditionFalse
```
nprokopic, line 58: - `newUpgradingCondition != nil`: `Upgrading` condition is set to some value,
- `newUpgradingCondition.Status != corev1.ConditionTrue && newUpgradingCondition.Status != corev1.ConditionFalse`: and that value is neither `True` nor `False`

`ConditionStatus` status is basically a string (with `True`, `False` and `Unknown` defined in core k8s package). We want to allow only `True` and `False`, and since we already blocked `Unknown` status (available as a const in upstream k8s core v1 package), here we block some random value that could be set.

internal/conditions/valdiatecluster.go
```diff
@@ -0,0 +1,122 @@
...
+		return microerror.Maskf(errors.InvalidConditionModificationError, errorMessageFormat, newUpgradingCondition.Status)
+	}
+
+	oldUpgradingCondition := capiconditions.Get(oldClusterCR, aeconditions.UpgradingCondition)
+
+	// Removing Upgrading condition is not allowed.
+	if oldUpgradingCondition != nil && newUpgradingCondition == nil {
+		errorMessage := "Removing Upgrading condition is not allowed"
+		return microerror.Maskf(errors.InvalidConditionModificationError, errorMessage)
+	}
```
nprokopic, line 71: - `oldUpgradingCondition != nil`: `Upgrading` condition was set in the CR
- `newUpgradingCondition == nil`: but now it has been removed

We don't allow removing of `Upgrading` condition, once it has been set, it can only change between `True` and `False`, depending on if the cluster is being upgraded or not.

internal/conditions/valdiatecluster.go
```diff
@@ -0,0 +1,122 @@
...
+		// release version with which the cluster was created, or release
+		// version to which the cluster was successfully upgraded the last
+		// time. Therefore, the release version set in label
+		// 'release.giantswarm.io/version', which represents the desired
+		// cluster version, must be either (1) same as the release version that
+		// is saved in condition, which is the state after cluster creation or
+		// cluster upgrade, or (2) greater than the release version that is
+		// saved in condition, which is state when the upgrade has been
+		// initiated.
+		if !desiredReleaseVersion.GTE(*lastSavedReleaseVersion) {
```
nprokopic, line 109: @fiunchinho the comment above the check hopefully explains it well.

pkg/azureupdate/validate_azureclusterconfig.go
```diff
@@ -62,20 +61,6 @@ func (a *AzureClusterConfigValidator) Validate(ctx context.Context, request *v1b
...
-		// The AzureClusterConfig CR doesn't have an indication of the fact that an update is in progress.
-		// I need to use the corresponding AzureConfig CR for this check.
-		acName := AzureClusterConfigOldCR.Spec.Guest.ID
-		ac := &v1alpha1.AzureConfig{}
-		err := a.ctrlClient.Get(ctx, client.ObjectKey{Name: acName, Namespace: AzureClusterConfigNewCR.Namespace}, ac)
-		if err != nil {
-			return microerror.Maskf(errors.InvalidOperationError, "Unable to find AzureConfig %s. Can't reliably tell if the cluster upgrade is safe or not. Error was %s", acName, err)
-		}
-
-		upgrading, status := clusterIsUpgrading(ac)
```
nprokopic, line 74: Removed the check of legacy `AzureConfig` conditions. cc @whites11

pkg/cluster/validate_update.go
```diff
@@ -120,3 +119,37 @@ func validateClusterNetworkUnchanged(old capiv1alpha3.Cluster, new capiv1alpha3.
...
+			forbidden = true
+		} else if capiconditions.IsTrue(clusterOldCR, aeconditions.UpgradingCondition) {
+			setCondition = aeconditions.UpgradingCondition
+			message = "cluster is already being upgraded"
+			forbidden = true
+		}
+
+		if forbidden {
+			return microerror.Maskf(errors.InvalidOperationError, "upgrade cannot be initiated now, Cluster condition %s is set to True, %s", setCondition, message)
+		}
```
nprokopic, line 151: Checking if the cluster is currently being created or upgraded when the upgrade is initiated. This is meant to replace previous `AzureConfig` conditions check. cc @whites11

internal/conditions/valdiate_cluster.go
```diff
@@ -0,0 +1,79 @@
+package conditions
```
tuommaki, line 1: Typo in filename: `s/valdiate/validate/`

internal/conditions/valdiate_cluster.go
```diff
@@ -0,0 +1,79 @@
...
+		validateRemovingConditionIsNotAllowed,
+	}
+
+	err = validateAll(oldCreatingCondition, newCreatingCondition, validations)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	creationWasCompleted := oldCreatingCondition != nil && oldCreatingCondition.Status == corev1.ConditionFalse
+	nowIsInCreation := newCreatingCondition != nil && newCreatingCondition.Status == corev1.ConditionTrue
```
tuommaki, line 47: Typo I presume? `s/now/new`

tuommaki (approved):
Few typos but otherwise very good. :+1:

internal/conditions/valdiate_cluster.go
nprokopic: no, it was on purpose :smile: the idea was to read it like "if creation was completed and now is in creation" :)

tuommaki: :see_no_evil:

nprokopic: :smile:

nprokopic: basically "if creation was completed and now is in creation again, what heck are you doing, you cannot restart cluster creation after it's done" :smile:

internal/conditions/validate.go
```diff
@@ -0,0 +1,70 @@
...
+		err = validation(oldCondition, newCondition)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	return nil
+}
+
+func validateNewStatusUnknownIsNotAllowed(_, newCondition *capi.Condition) error {
```
fiunchinho, line 26: isn't this validation redundant because `validateNewStatusMustBeEitherTrueOrFalse` already validates is either true or false?

internal/conditions/validate_cluster.go
```diff
@@ -0,0 +1,79 @@
...
+		validateNewStatusMustBeEitherTrueOrFalse,
+		validateRemovingConditionIsNotAllowed,
+	}
+
+	err = validateAll(oldCreatingCondition, newCreatingCondition, validations)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	creationWasCompleted := oldCreatingCondition != nil && oldCreatingCondition.Status == corev1.ConditionFalse
```
fiunchinho, line 46: couldn't this be validated like the other funcs?

internal/conditions/validate_cluster.go
```diff
@@ -0,0 +1,79 @@
...
+	nowIsInCreation := newCreatingCondition != nil && newCreatingCondition.Status == corev1.ConditionTrue
+
+	// Rule: Creating cannot be changed from False to True.
+	// Why: This change would mean that the cluster creation has started again,
+	//      which is not possible.
+	if creationWasCompleted && nowIsInCreation {
+		const errorMessage = "Modifying Creating condition from False to True is not allowed, " +
+			"as that would mean that the cluster creation is started again, which is not possible"
+		return microerror.Maskf(errors.InvalidConditionModificationError, errorMessage)
+	}
```
fiunchinho, line 56: couldn't this be validated like the other funcs?

fiunchinho (commented):
small comments, but looks ready!

internal/conditions/validate_cluster.go
nprokopic: This is specific for `Creating` condition, while validation functions in `validate.go` are kind of generic and can be used to validate different conditions.

Or you just meant to put this into a separate function? Maybe like a generic check of transition from `False` to `True`.

fiunchinho: Nah, I kind of like how explicit is now, and by using a function we may lose that. So let's keep it like it is.

pkg/azurecluster/validate_update.go
```diff
@@ -67,18 +68,38 @@ func (a *UpdateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
...
-func (a *UpdateValidator) Log(keyVals ...interface{}) {
-	a.logger.Log(keyVals...)
+		clusterCRReleaseVersion, err := semverhelper.GetSemverFromLabels(cluster.Labels)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if !newClusterVersion.Equals(clusterCRReleaseVersion) {
+			return microerror.Maskf(errors.InvalidOperationError, "AzureCluster release version must be set to the same release version as Cluster CR release version label")
+		}
```
nprokopic, line 101: @fiunchinho we blocked azure-operator from upgrading `AzureCluster` :see_no_evil: I changed this to a simple check that the new `AzureCluster` release simply matches `Cluster` release.

fiunchinho: Ok, that makes sense.

nprokopic: Works now, `AzureCluster` release label is updated, upgrade is in progress.

internal/conditions/validate.go
nprokopic: It does, but I added a separate validation for `Unknown` on purpose, as `Unknown` status is defined in k8s core package and could be set more easily and on purpose, while we don't want that.

Other possible values, other than `Unknown`, `True` and `False`, could be some new status values in some new upstream package versions that we did not take into account yet, or simply a mistake somewhere in code (which could be also customer's code now with CP API access).

fiunchinho: ok, it makes sense :+1:

nprokopic: So basically distinguishing between `Unknown` that we know about and other unknown status values :)

PR #60, opened 2020-10-28
Validate that organization label contains existing organization

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/14010

I'm checking the organization label for most of the CRs. This means that all tests would fail unless the default organization was created, so I did that.

Question: should we validate `AzureConfig` or `AzureClusterConfig`?


pkg/generic/validate_organization_label.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	if old.GetLabels()[label.Organization] != new.GetLabels()[label.Organization] {
+		return microerror.Maskf(errors.InvalidOperationError, "Organization label %#q must not be changed", label.Organization)
+	}
+
+	return nil
+}
+
+func ValidateOrganizationLabelContainsExistingOrganization(ctx context.Context, obj metav1.Object, ctrlClient client.Client) error {
```
fiunchinho, line 31: This is the meat.

fiunchinho:
Tested in `godsmack`

pkg/generic/validate_organization_label.go
tuommaki: Minor nitpick, no strong feelings, but my intuition would put `ctrlClient` before the `obj` in parameter ordering. I cannot explain why, but it's just a feeling. That being said, it's just an aesthetic thing so completely your call :slightly_smiling_face:

pkg/generic/validate_organization_label.go
```diff
@@ -0,0 +1,47 @@
...
+	organizationName, ok := obj.GetLabels()[label.Organization]
+	if !ok {
+		return microerror.Maskf(errors.NotFoundError, "CR doesn't contain Organization label %#q", label.Organization)
+	}
+
+	organization := &securityv1alpha1.Organization{}
+	err := ctrlClient.Get(ctx, client.ObjectKey{Name: organizationName}, organization)
```
tuommaki, line 39: Does this work with mixed-case names? I mean, k8s object cannot contain uppercase letters (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) - should it be normalized?

tuommaki (changes requested):
Normalization might be missing

fiunchinho:
Added normalization, please take a look.

pkg/generic/validate_organization_label.go
```diff
@@ -0,0 +1,48 @@
...
+func ValidateOrganizationLabelContainsExistingOrganization(ctx context.Context, ctrlClient client.Client, obj metav1.Object) error {
+	organizationName, ok := obj.GetLabels()[label.Organization]
+	if !ok {
+		return microerror.Maskf(errors.NotFoundError, "CR doesn't contain Organization label %#q", label.Organization)
+	}
+
+	organization := &securityv1alpha1.Organization{}
+	err := ctrlClient.Get(ctx, client.ObjectKey{Name: normalize.AsDNSLabelName(organizationName)}, organization)
+	if apierrors.IsNotFound(err) {
+		return microerror.Maskf(errors.InvalidOperationError, "Organization label %#q must contain an existing organization, got %#q", label.Organization, organizationName)
```
tuommaki, line 42:
```suggestion
		return microerror.Maskf(errors.InvalidOperationError, "Organization label %#q must contain an existing organization, got %#q but didn't find any CR with name %#q", label.Organization, organizationName, normalize.AsDNSLabelName(organizationName))
```

tuommaki (approved):

PR #64, opened 2020-10-29
Simplify Validator interface

fiunchinho:
Until now our `Validator` interface was returning `(bool, error)`. Looking at the code, it seems like every time there is an error, we `return false, err`, otherwise we `return true, nil`. So I believe it's simpler if we get rid of the boolean value and just rely on the existence of the error: if there was an error the request was not allowed, otherwise it's allowed.

WDYT?

Most of the changes were made by IntelliJ so don't worry about telling me it's an stupid idea, it was a rather quick change.


whites11 (approved):
I completely agree

PR #65, opened 2020-10-29
MachinePool default Spec.Replicas

nprokopic:
Setting default value for `Spec.Replicas` to 1, as specified in Cluster API.

Default value is set on:
- `MachinePool` create, when `Spec.Replicas` is not specified,
- `MachinePool` update, when `Spec.Replicas` field has been set to `nil` (so we set it back to the default value).

### DIFF
--- a/pkg/machinepool/defaults.go
+++ b/pkg/machinepool/defaults.go
@@ -3,7 +3,6 @@ package machinepool
 import (
 	"context"
 	"fmt"
-	"strconv"
 
 	capiexp "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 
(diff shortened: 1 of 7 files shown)


pkg/machinepool/mutate_machinepool_create.go
```diff
@@ -39,11 +39,16 @@ func (m *CreateMutator) Mutate(ctx context.Context, request *v1beta1.AdmissionRe
 		return result, nil
 	}
 
-	machinePoolCR := &capiv1alpha3.Cluster{}
+	machinePoolCR := &capiexp.MachinePool{}
```
whites11, line 42: Was this a bug on master?

whites11 (approved):
LGTM when tested

pkg/machinepool/mutate_machinepool_create.go
nprokopic: yup :)

pkg/machinepool/defaults.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	return patches
+}
+
+// setDefaultReplicaValue checks if Spec.Replicas has been set, and if it is
+// not, it sets its value to 1.
+func (m *CreateMutator) setDefaultReplicaValue(ctx context.Context, machinePool *capiexp.MachinePool) *mutator.PatchOperation {
+	if machinePool.Spec.Replicas == nil {
+		const defaultReplicas = "1"
```
fiunchinho, line 29: We normally define the constants in the top of the file, but I'm not blocking. Your call.

fiunchinho (approved):

fiunchinho (changes requested):
Update the changelog please

pkg/machinepool/defaults.go
nprokopic: I basically needed a local var, which is not used anywhere else, but then changed it to `const` without too much thinking, as it does not change nor it should change.

nprokopic:
> Update the changelog please

Updated.

fiunchinho (approved):

PR #66, opened 2020-11-03
Add integration test

fiunchinho:
This PR adds a simple integration test using `apptestctl`. The test is really simple but just by running we are testing that the application compiles and the helm chart is deployed fine.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,7 @@ workflows:
           name: admission-integration-test
           install-app-platform: true
           test-dir: "integration/test/organization"
+          test-timeout: "10m"
           filters:
             # Do not trigger the job on merge to master.
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,3 +58,15 @@ workflows:
             tags:
               only: /^v.*/
 
+      - architect/integration-test:
+          context: architect
+          name: admission-integration-test
+          setup-script: "integration/test/organization/setup.sh"
+          test-dir: "integration/test/organization"
+          filters:
+            # Do not trigger the job on merge to master.
+            branches:
+              ignore:
+                - master
+          requires:
+            - push-azure-admission-controller-to-control-plane-app-catalog
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@0.14.0
+  architect: giantswarm/architect@0.17.0
 
 
 workflows:
@@ -61,7 +61,7 @@ workflows:
       - architect/integration-test:
           context: architect
           name: admission-integration-test
-          setup-script: "integration/test/organization/setup.sh"
+          install-app-platform: true
           test-dir: "integration/test/organization"
           filters:
             # Do not trigger the job on merge to master.
--- a/go.mod
+++ b/go.mod
@@ -10,11 +10,11 @@ require (
 	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/ghodss/yaml v1.0.0
 	github.com/giantswarm/apiextensions/v2 v2.6.2
-	github.com/giantswarm/apiextensions/v3 v3.6.0
+	github.com/giantswarm/apiextensions/v3 v3.7.0
 	github.com/giantswarm/apptest v0.5.0
 	github.com/giantswarm/k8sclient/v4 v4.0.0
 	github.com/giantswarm/microerror v0.2.1
-	github.com/giantswarm/micrologger v0.3.3
+	github.com/giantswarm/micrologger v0.3.4
 	github.com/google/go-cmp v0.5.2
 	github.com/stretchr/testify v1.6.1
 	golang.org/x/tools v0.0.0-20200706234117-b22de6825cf7 // indirect
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/giantswarm/apiextensions/v2 v2.6.2
 	github.com/giantswarm/apiextensions/v3 v3.6.0
-	github.com/giantswarm/apptest v0.4.2-0.20201103172001-0ef9f68e6299
+	github.com/giantswarm/apptest v0.5.0
 	github.com/giantswarm/k8sclient/v4 v4.0.0
 	github.com/giantswarm/microerror v0.2.1
 	github.com/giantswarm/micrologger v0.3.3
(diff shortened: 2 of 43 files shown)


go.mod
```diff
@@ -10,7 +10,8 @@ require (
 	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/ghodss/yaml v1.0.0
 	github.com/giantswarm/apiextensions/v2 v2.6.2
-	github.com/giantswarm/apiextensions/v3 v3.4.1
+	github.com/giantswarm/apiextensions/v3 v3.6.0
+	github.com/giantswarm/apptest v0.4.2-0.20201103172001-0ef9f68e6299
```
fiunchinho, line 14: I will change this to use a proper release when we release next `apptest` version.

integration/test/organization/main_test.go
```diff
@@ -0,0 +1,137 @@
...
+	prodCatalogUrl  = "https://giantswarm.github.io/control-plane-catalog"
+	// API Groups for upstream Cluster API types.
+	clusterAPIGroup                    = "cluster.x-k8s.io"
+	infrastructureAPIGroup             = "infrastructure.cluster.x-k8s.io"
+	experimentalClusterAPIGroup        = "exp.cluster.x-k8s.io"
+	experimentalInfrastructureAPIGroup = "exp.infrastructure.cluster.x-k8s.io"
+)
+
+var (
+	appTest apptest.Interface
```
tuommaki, line 35: Do we need to make this instance global variable? Is there something in the `apptest` design or implementation that requires it or makes it convenient to have it that way? I noticed that tests are using shared client from that instance and it brings high risk of subtle bugs in tests due to overlapping CRs. When dealing with CRDs it quickly becomes tedious and we often make helper functions for that, but this often then leads to shared CR names / IDs because normally those don't matter but in this specific case it's suddenly very important aspect.

rossf7: The apptest library is still pretty fresh. One of the goals is to have simple wiring in components like metrics-server where we only use Go for the integration tests.

The clients are a convenience wrapper for that. But especially with the controller client I can see how it could cause problems. They don't have to be used and could even be removed. All that is required is a kubeconfig.

integration/test/organization/main_test.go
```diff
@@ -0,0 +1,137 @@
...
+			KubeConfigPath: env.KubeConfig(),
+			Logger:         logger,
+		})
+		if err != nil {
+			panic(err.Error())
+		}
+	}
+
+	{
+		values := `Installation:
```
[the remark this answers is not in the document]
rossf7: If this can be done in a lightweight way I think this would be nice.

My only concern is from past experience with the e2etemplates library. Some of those templates got complex and needing to update multiple repos slows down working on tests.

integration/test/organization/main_test.go
```diff
@@ -0,0 +1,137 @@
...
+package organization
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	corev1alpha1 "github.com/giantswarm/apiextensions/v2/pkg/apis/core/v1alpha1"
+	providerv1alpha1 "github.com/giantswarm/apiextensions/v2/pkg/apis/provider/v1alpha1"
```
rossf7, line 12:
```suggestion
	corev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/core/v1alpha1"
	providerv1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/provider/v1alpha1"
```

Can we use v3 for all of these?

integration/test/organization/main_test.go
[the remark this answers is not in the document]
tuommaki: Yeah, I understand that, but here it would be in the same repo so that wouldn't be so much of a problem. I've been recently using a lot of pattern where unit test input CRs are stored as file and loaded during the test and I've personally liked that a lot. Removes a bit of boilerplate from the test code. See e.g. https://github.com/giantswarm/azure-operator/blob/master/integration/test/crmapping/crmapping_test.go#L110-L177

integration/test/organization/main_test.go
tuommaki: We talked with Jose about this and when the test is performed towards shared k8s API due to nature of admission controller, the shared client can't be avoided in the end. We just need to be careful when writing tests that we don't share IDs between them :slightly_smiling_face:  All good here. :+1:

integration/test/organization/main_test.go
fiunchinho: Fixed

integration/test/organization/main_test.go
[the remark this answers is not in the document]
fiunchinho: I tried the template thingy for the values. Take a look @tuommaki

.circleci/config.yml
```diff
@@ -58,3 +58,15 @@ workflows:
             tags:
               only: /^v.*/
 
+      - architect/integration-test:
+          context: architect
+          name: admission-integration-test
+          setup-script: "integration/test/organization/setup.sh"
```
rossf7, line 64:
```suggestion
          install-app-platform: true
```

If you use the latest architect-orb (0.17.0) then you don't need the setup script.

We need docs for that. I'll work on them this week.

integration/env/common.go
```diff
@@ -0,0 +1,92 @@
...
+	kubeconfig = os.Getenv(EnvVarE2EKubeconfig)
+	if kubeconfig == "" {
+		panic(fmt.Sprintf("env var '%s' must not be empty", EnvVarE2EKubeconfig))
+	}
+
+	kubeconfig = os.Getenv(EnvVarE2EKubeconfig)
+	if kubeconfig == "" {
+		panic(fmt.Sprintf("env var '%s' must not be empty", EnvVarE2EKubeconfig))
+	}
+
```
rossf7, line 48:
```suggestion
```

This part is repeated.

rossf7 (approved):
LGTM. Just a couple of minor things.

integration/template/values.go
```diff
@@ -0,0 +1,53 @@
+package template
```
tuommaki, line 1: `template.Values()` somehow doesn't connect with `values.yaml` for me. What if this package was called `values` and `Values()` was `YAML()`?

The template itself below is naturally `ValuesTemplate` which would kind of support `values` package and it could potentially be renamed to just `Template` even.

The logic related to text template rendering is often so small that I wouldn't maybe name a package after it when there's no other major logic related to templating.

WDYT?

fiunchinho: Done PTAL

tuommaki (approved):
:rocket:

PR #72, opened 2020-11-05
Allow empty string as failureDomain for azureMachine

whites11:
This PR allows to specify an empty string as failure domain.
That means the availability zone selection happens automatically.
This is what happa does (passes "" and not nil).

### DIFF
--- a/pkg/azuremachine/failure_domain.go
+++ b/pkg/azuremachine/failure_domain.go
@@ -10,7 +10,7 @@ import (
 
 func validateFailureDomain(azureMachine capzv1alpha3.AzureMachine, supportedAZs []string) error {
 	// No failure domain specified.
-	if azureMachine.Spec.FailureDomain == nil {
+	if azureMachine.Spec.FailureDomain == nil || *azureMachine.Spec.FailureDomain == "" {
 		return nil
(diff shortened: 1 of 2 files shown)


pkg/azuremachine/failure_domain.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func validateFailureDomain(azureMachine capzv1alpha3.AzureMachine, supportedAZs []string) error {
 	// No failure domain specified.
-	if azureMachine.Spec.FailureDomain == nil {
+	if azureMachine.Spec.FailureDomain == nil || *azureMachine.Spec.FailureDomain == "" {
```
fiunchinho, line 13: I was thinking if it would help our future selves to use a constant called `AutomaticFailureDomainSelection` or something like that for the empty value.

whites11: but what about nil?
It would be needed to have 2 constants, because nil and "" mean the same thing in this scenario

fiunchinho: Ok, never mind.

fiunchinho (approved):

PR #73, opened 2020-11-05
Relax failuredomain immutability check

whites11:
This PR allows changing the `AzureMachine`'s FaultDomain field from nil to empty string and vice versa.
The meaning of the two values is identical so this makes the validation process less strict while being equally valid.

### DIFF
--- a/pkg/azuremachine/failure_domain.go
+++ b/pkg/azuremachine/failure_domain.go
@@ -35,7 +35,7 @@ func validateFailureDomainUnchanged(old capzv1alpha3.AzureMachine, new capzv1alp
 		return nil
 	}
 
-	// Allow changing from nil to "" and from "" to nil. They are synonims.
+	// Allow changing from nil to "" and from "" to nil. They are synonyms.
 	if old.Spec.FailureDomain == nil && *new.Spec.FailureDomain == "" ||
 		*old.Spec.FailureDomain == "" && new.Spec.FailureDomain == nil {
 		return nil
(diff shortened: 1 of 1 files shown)


nprokopic:
Where do we have a need to support an empty string (in addition to `nil`)?

IMHO it would be simpler if we just accepted either `nil` or a supported Failure Domain value (valid AZ number), it would simplify the operator code that uses the field and then talks to Azure. If both `nil` and `""` are possible, then wherever the field is used in code, both checks should be probably done in order to avoid errors (as opposed to checking only for nil, and trusting non-nil values to be valid AZ numbers).

E.g. I took a look at CAPZ code, there are simple ` != nil` checks in few places, so I guess when those checks pass, it's probably expected that correct FailureDomain/AZ value is set.

Not blocking though.

nprokopic:
Ok, just saw in this PR https://github.com/giantswarm/azure-admission-controller/pull/72 that happa sends `""`. @axbarsan could happa not set `""` and just leave the value to be `nil`?

whites11:
it's the azure-operator that switches from "" to nil

pkg/azuremachine/failure_domain.go
```diff
@@ -35,6 +35,12 @@ func validateFailureDomainUnchanged(old capzv1alpha3.AzureMachine, new capzv1alp
 		return nil
 	}
 
+	// Allow changing from nil to "" and from "" to nil. They are synonims.
```
tuommaki, line 38:
```suggestion
	// Allow changing from nil to "" and from "" to nil. They are synonyms.
```

tuommaki (approved):

PR #74, opened 2020-11-06
Copy release

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/14156

This PR ensures the `release.giantswarm.io/version` label is set for `MachinePool` and `AzureMachinePool` CRs at creation time if not set.
It takes the value from the `Cluster` CR.


tuommaki:
Hmm... Now that I'm thinking, is this necessarily needed at all? Theoretically in NP-to-NP upgrade PR, the `clusterupgrade` handler should always propagate latest release & azure-operator version labels to MachinePools that belong to that given cluster. `machinepoolupgrade` handler would then enforce the same propagation from `MachinePool` to corresponding `AzureMachinePool`. Maybe we should test this use-case with the the branch once these few other ones are done?

whites11:
> Hmm... Now that I'm thinking, is this necessarily needed at all? Theoretically in NP-to-NP upgrade PR, the `clusterupgrade` handler should always propagate latest release & azure-operator version labels to MachinePools that belong to that given cluster. `machinepoolupgrade` handler would then enforce the same propagation from `MachinePool` to corresponding `AzureMachinePool`. Maybe we should test this use-case with the the branch once these few other ones are done?

Should it work on node pool creation as well?

tuommaki:
> Should it work on node pool creation as well or during an upgrade only?

The Cluster -> MachinePool logic I need to carefully verify, but the MachinePool -> AzureMachinePool is unconditional, it should always work.

whites11:
> > Should it work on node pool creation as well or during an upgrade only?
> 
> The Cluster -> MachinePool logic I need to carefully verify, but the MachinePool -> AzureMachinePool is unconditional, it should always work.

ok I will test it now.

tuommaki:
> ok I will test it now.

You need to use the version from branch `machinepool-add-upgrading-condition` for that. There is release on `ghost`: `v42.0.1`.

I checked the code and when compiling/running it in my head, it works on creation as well.

whites11:
> > ok I will test it now.
> 
> You need to use the version from branch `machinepool-add-upgrading-condition` for that. There is release on `ghost`: `v42.0.1`.
> 
> I checked the code and when compiling/running it in my head, it works on creation as well.

Tuomas the human compiler

whites11:
In general Tuomas I don't know if I like the idea of leaving the population of labels to Azure Operator.
We used to set all those labels "in advance" and we should keep doing like that.
I feel like we could one day remove or change the label duplication feature from azure operator for whatever reason and the cluster creation will stop working.
This change here is reasonably simple and it solves the problem once and for all simply by keeping things as they were.

WDYT?

pkg/generic/ensure_release_label_test.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	if release != nil {
+		obj.Labels[label.ReleaseVersion] = *release
+	}
+
+	return obj
+}
+
+//func Test_WhenCreatingClusterWithExistingOrganizationThenValidationSucceeds(t *testing.T) {
```
tuommaki, line 145: If this is not needed, please delete it so it doesn't confuse anyone in the future :slightly_smiling_face:

tuommaki (approved):
I'm fine either way. I was just bringing that up in case we want to keep the admission controller as small as possible, but this is straightforward in general so I'm fine with this. Do as you see best :+1:

whites11:
BTW I tested it with the az op version you mentioned and worker nodes never came up.

PR #78, opened 2020-11-09
Use builder to build AzureCluster during tests

fiunchinho:
This PR introduces a builder so that we don't have a single function that accepts all the potential parameters required to customize an `AzureCluster`. Having a single function forces us to pass all parameters every time we use it, which adds noise to the tests, i.e tests that are related to the location field need to know about the API endpoint fields.

Using the builder, each test only passes the parameters it needs to customize the `AzureCluster` object, and all other fields get a sensible default.

Before applying the pattern to all the tests I want to gather your feedback. WDYT?

### DIFF
--- a/internal/test/azurecluster/builder.go
+++ b/internal/test/azurecluster/builder.go
@@ -2,13 +2,26 @@ package azurecluster
 
 import (
 	"encoding/json"
+	"math/rand"
 
+	"github.com/giantswarm/apiextensions/v3/pkg/label"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
-	"sigs.k8s.io/cluster-api/api/v1alpha3"
+	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
 )
 
 type BuilderOption func(azureCluster *capzv1alpha3.AzureCluster) *capzv1alpha3.AzureCluster
+
+var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+func Name(name string) BuilderOption {
+	return func(azureCluster *capzv1alpha3.AzureCluster) *capzv1alpha3.AzureCluster {
+		azureCluster.ObjectMeta.Name = name
+		azureCluster.Labels[capiv1alpha3.ClusterLabelName] = name
+		azureCluster.Labels[label.Cluster] = name
+		return azureCluster
+	}
+}
 
 func Location(location string) BuilderOption {
 	return func(azureCluster *capzv1alpha3.AzureCluster) *capzv1alpha3.AzureCluster {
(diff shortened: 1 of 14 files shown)


pkg/azurecluster/builder.go
```diff
@@ -0,0 +1,68 @@
+package azurecluster
```
tuommaki, line 1: Since this is tightly coupled with testing, should something related be set as in-between package? For example `pkg/test/azurecluster` ? It is a bit unfortunate that language already allocated `testing` which is mandatory pretty much always, but but maybe the singular form works?

pkg/azurecluster/builder.go
```diff
@@ -0,0 +1,68 @@
...
+
+func ControlPlaneEndpoint(controlPlaneEndpointHost string, controlPlaneEndpointPort int32) BuilderOption {
+	return func(azureCluster *capzv1alpha3.AzureCluster) *capzv1alpha3.AzureCluster {
+		azureCluster.Spec.ControlPlaneEndpoint.Host = controlPlaneEndpointHost
+		azureCluster.Spec.ControlPlaneEndpoint.Port = controlPlaneEndpointPort
+		return azureCluster
+	}
+}
+
+func BuildAzureCluster(clusterName string, opts ...BuilderOption) *capzv1alpha3.AzureCluster {
```
tuommaki, line 28: Proposal: In order to avoid unintentional conflicts between tests, would it make sense to change this slightly towards randomized CR naming:
```suggestion
func BuildAzureCluster(opts ...BuilderOption) *capzv1alpha3.AzureCluster {
```

Here, the sane default would be random string generated on function all and if test wants to specify the object name, it could be done with a `BuilderOption()`.

tuommaki:
Oh, to kind of state explicitly my general feeling towards this: I like this. Should be a big improvements towards writing tests. :+1:

pkg/azurecluster/builder.go
fiunchinho: what about `internal/test/azurecluster`?

pkg/azurecluster/builder.go
fiunchinho: I like it!

pkg/azurecluster/builder.go
tuommaki: Works for me. :+1:

fiunchinho:
Added builders for `AzureCluster`, `MachinePool` and `AzureMachinePool`. We can add more builders later on.

internal/test/azurecluster/builder.go
```diff
@@ -0,0 +1,90 @@
...
+
+func BuildAzureClusterAsJson(opts ...BuilderOption) []byte {
+	azureCluster := BuildAzureCluster(opts...)
+
+	byt, _ := json.Marshal(azureCluster)
+
+	return byt
+}
+
+func generateName() string {
```
tuommaki, line 84: Would it work if this was moved to `internal/test` and used from there in all these sub-packages?

internal/test/azuremachinepool/builder.go
```diff
@@ -0,0 +1,131 @@
...
+
+func BuildAzureMachinePool(opts ...BuilderOption) *expcapzv1alpha3.AzureMachinePool {
+	nodepoolName := generateName()
+	azureMachinePool := &expcapzv1alpha3.AzureMachinePool{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nodepoolName,
+			Namespace: "org-giantswarm",
+			Labels: map[string]string{
+				label.AzureOperatorVersion: "5.0.0",
+				label.Cluster:              "ab123",
```
tuommaki, line 78: Maybe add `capiv1alpha3.ClusterLabelName` here and below (`MachinePool`) as well?

tuommaki (approved):
LGTM, minor notes, do as you see best :slightly_smiling_face:

PR #79, opened 2020-11-09
Ensure operator version labels

whites11:
Towards https://github.com/giantswarm/giantswarm/issues/14156

This PR ensures the azure-operator and cluster-operator labels are set on create in the MachinePool and AzureMachinePool CRs.
This removes the need to know the release version on kubectl-gs when creating a node pool.

### DIFF
--- a/pkg/generic/ensure_component_version_label.go
+++ b/pkg/generic/ensure_component_version_label.go
@@ -12,7 +12,7 @@ import (
 	"github.com/giantswarm/azure-admission-controller/pkg/mutator"
 )
 
-func EnsureComponentVersionLabel(ctx context.Context, ctrlClient client.Client, meta metav1.Object, componentName string, labelName string) (*mutator.PatchOperation, error) {
+func EnsureComponentVersionLabel(ctx context.Context, ctrlClient client.Client, meta metav1.Object, labelName string) (*mutator.PatchOperation, error) {
 	if meta.GetLabels()[labelName] == "" {
 		componentVersion, err := getLabelValueFromCluster(ctx, ctrlClient, meta, labelName)
 		if err != nil {
--- a/pkg/generic/ensure_component_version_label.go
+++ b/pkg/generic/ensure_component_version_label.go
@@ -3,12 +3,8 @@ package generic
 import (
 	"context"
 	"fmt"
-	"strings"
 
-	releasev1alpha1 "github.com/giantswarm/apiextensions/v2/pkg/apis/release/v1alpha1"
-	"github.com/giantswarm/apiextensions/v3/pkg/label"
 	"github.com/giantswarm/microerror"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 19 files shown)


fiunchinho:
Doesn't `Cluster` contain the components' labels? I was wondering if it would be easier to just read the versions from the labels in the `Cluster` CR like you do to find out the Release version, instead of fetching the `Release`.

tuommaki:
Why `cluster-operator` version is added to CAPI or CAPZ CRs? The version we can use doesn't know anything about those types. It only reconciles `AzureClusterConfig` CR. I think it shouldn't be there. It doesn't do anything effectively, but creates noise in the logs.

whites11:
> Doesn't `Cluster` contain the components' labels? I was wondering if it would be easier to just read the versions from the labels in the `Cluster` CR like you do to find out the Release version, instead of fetching the `Release`.

And fail if they're not there?
Might be a good idea.

whites11:
> Why `cluster-operator` version is added to CAPI or CAPZ CRs? The version we can use doesn't know anything about those types. It only reconciles `AzureClusterConfig` CR. I think it shouldn't be there. It doesn't do anything effectively, but creates noise in the logs.

Can't answer for sure.
`kubectl-gs` was doing that and I mimicked the same exact behaviour

whites11:
@fiunchinho pushed the change to meet your suggestion

pkg/generic/ensure_component_version_label_test.go
```diff
@@ -0,0 +1,206 @@
...
+			fakeK8sClient := unittest.FakeK8sClient()
+			ctrlClient := fakeK8sClient.CtrlClient()
+
+			// Cluster with both operator annotations.
+			ab123 := &v1alpha3.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "ab123",
+					Namespace: "default",
+					Labels: map[string]string{
+						//"release.giantswarm.io/version": "13.0.0",
```
tuommaki, line 108: Needed or not?
```suggestion
```

tuommaki (approved):

pkg/generic/ensure_release_label.go
```diff
@@ -48,6 +40,30 @@ func EnsureReleaseVersionLabel(ctx context.Context, ctrlClient client.Client, me
 	return nil, nil
 }
 
+func getLabelValueFromCluster(ctx context.Context, ctrlClient client.Client, meta metav1.Object, labelName string) (string, error) {
+	// Get version from cluster.
```
fiunchinho, line 44:
```suggestion
```

pkg/generic/ensure_component_version_label.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-admission-controller/internal/errors"
+	"github.com/giantswarm/azure-admission-controller/pkg/mutator"
+)
+
+func EnsureComponentVersionLabel(ctx context.Context, ctrlClient client.Client, meta metav1.Object, componentName string, labelName string) (*mutator.PatchOperation, error) {
```
fiunchinho, line 15: `componentName` is not required anymore.

PR #82, opened 2020-11-10
Added documentation about managed CRs and fields

whites11:
This PR adds 2 documentation pages that list the CRs and fields that are considered by the azure admission controller


tuommaki (approved):
Beautiful!

PR #84, opened 2020-11-11
Use specific errors for specific business rules

fiunchinho:
We were using a common error object to signal validation errors for different business rules. I wanted to use a specific error for every business rule so that tests can't accidentally work.


whites11 (approved):
Looking quickly I didn't find any obvious problems

tuommaki (approved):
LGTM

PR #94, opened 2020-11-16
Fix validation for alpha releases, add possibility to ignore a release

nprokopic:
What's in the box?

- Skip alpha releases when checking if major and minor releases were skipped for upgrades.
- Skip validating releases that have annotation `release.giantswarm.io/ignore: "true"` (they only have to exist)


internal/releaseversion/releaseversion.go
```diff
@@ -12,21 +12,30 @@ import (
 	"github.com/giantswarm/azure-admission-controller/internal/errors"
 )
 
+const (
+	ignoreReleaseAnnotation = "release.giantswarm.io/ignore"
```
tuommaki, line 16: Maybe copy this to `apiextensions` as well?

internal/releaseversion/releaseversion.go
```diff
@@ -56,23 +68,57 @@ func Validate(ctx context.Context, ctrlCLient client.Client, oldVersion semver.V
...
 		}
 		ret = append(ret, &parsed)
+		releases = append(releases, release)
 	}
 
-	return ret, nil
+	return ret, releases, nil
+}
+
+func filterOutAlphaAndIgnoredReleases(releases []*semver.Version, releaseCRs []v1alpha1.Release) ([]*semver.Version, []v1alpha1.Release) {
```
tuommaki, line 92: Here, similar to `isOldOrNewReleaseIgnored()` below:

It seems that `releases` and `releaseCRs` must be sorted in the same order. Could `releases` be dropped and `releaseCRs[i].Name` be used instead? In this function only string of release version is needed and below the `semver` representation could be generated on-the-fly from `Release` object.

WDYT?

internal/releaseversion/releaseversion.go
nprokopic: yeah, I just wanted to run the idea with you folks first and try it out :)

internal/releaseversion/releaseversion.go
nprokopic: Makes sense :+1:

nprokopic: > It seems that releases and releaseCRs must be sorted in the same order. 

They are, as they are created in the same function at the same time.

nprokopic: Ok, I took a look now, and some parts of the validation are using `semver.Version` objects, needed for release version comparison (greater than, less than, so actual `Version` struct is needed).

I added `Release` CR there so that annotation can be checked for ignored releases.

I initially left usage of both slices on purpose so that we don't have to parse version name to `semver.Version` object multiple times.

internal/releaseversion/releaseversion.go
```diff
@@ -88,3 +137,12 @@ func included(releases []*semver.Version, release semver.Version) bool {
 func isAlphaRelease(release string) bool {
 	return strings.Contains(release, "alpha")
 }
+
+func isIgnoredRelease(releaseCR *v1alpha1.Release) bool {
+	ignoreValue, isIgnoreAnnotationSet := releaseCR.Annotations[ignoreReleaseAnnotation]
```
fiunchinho, line 142: I'm not sure if it even can happen, but should we check if `Annotations` is not nil?

nprokopic: Not sure, good point :+1: :fist_oncoming:

nprokopic: It does not break for `Annotations==nil`, it just returns `isIgnoreAnnotationSet` as `false`, example https://play.golang.org/p/2n40nlvAcby

nprokopic: We have too many `null` traumas from other languages :smile:

internal/releaseversion/spec.go
```diff
@@ -0,0 +1,11 @@
+package releaseversion
+
+import (
+	"github.com/blang/semver"
+	"github.com/giantswarm/apiextensions/v2/pkg/apis/release/v1alpha1"
+)
+
+type release struct {
```
nprokopic, line 8: @tuommaki I added this internal struct so there are no two slices that are passed around (`[]semver.Version` and `[]v1alpha1.Release`), but one with both info bundled together so each can be used where needed.

tuommaki (approved):
LGTM

fiunchinho (approved):

PR #105, opened 2020-12-04
validate organization for node pools

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/14519

this PR validates that machinepool and azuremachinepools CRs have the right organization (has to match their Cluster's) on create.

### DIFF
--- a/pkg/generic/error.go
+++ b/pkg/generic/error.go
@@ -13,13 +13,13 @@ func IsClusterNotFoundError(err error) bool {
 	return microerror.Cause(err) == clusterNotFoundError
 }
 
-var invalidOrganizationError = &microerror.Error{
-	Kind: "invalidOrganizationError",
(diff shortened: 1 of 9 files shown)


pkg/generic/error.go
```diff
@@ -4,6 +4,24 @@ import (
...
+var clusterNotFoundError = &microerror.Error{
+	Kind: "clusterNotFoundError",
+}
+
+// IsClusterNotFoundError asserts clusterNotFoundError.
+func IsClusterNotFoundError(err error) bool {
+	return microerror.Cause(err) == clusterNotFoundError
+}
+
+var invalidOrganizationError = &microerror.Error{
```
fiunchinho, line 16: maybe let's get a bit more specific here? What about `nodepoolOrgDoesNotMatchClusterOrgError`?

whites11: makese sense, done

fiunchinho (approved):

PR #134, opened 2021-01-11
Bump cert-manager apiversion to v1

njuettner:
We need to update cert manager to v1 otherwise we might also run into issues like kiam
see giantswarm/kiam-app#63

CPs have cert-manager-app-2.3.0 (cert-manager v1.0.2), so GA APIs can be used.


whites11 (approved):
LGTM

PR #135, opened 2021-01-14
Ignore Deprecated releases during upgrade validations.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/15142

This PR ignores deprecated releases when checking if an upgrade path is valid or not.

### DIFF
--- a/internal/releaseversion/releaseversion.go
+++ b/internal/releaseversion/releaseversion.go
@@ -2,7 +2,6 @@ package releaseversion
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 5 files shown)


internal/releaseversion/releaseversion.go
```diff
@@ -31,11 +32,6 @@ func Validate(ctx context.Context, ctrlCLient client.Client, oldVersion semver.V
 		return microerror.Maskf(releaseNotFoundError, "release %s was not found in this installation", newVersion)
 	}
 
-	// Skip validations for ignored releases.
```
whites11, line 34: This is dangerous, because we would allow ANY upgrade path for clusters running an ignored release. Even downgrading

pkg/azureupdate/validate_azureconfig_test.go
```diff
@@ -282,6 +321,43 @@ func TestAzureConfigValidate(t *testing.T) {
...
+			oldVersion:   "12.0.0",
+			newVersion:   "13.1.0",
+			errorMatcher: releaseversion.IsSkippingReleaseError,
+		},
+		{
+			name: "case 16",
+			ctx:  context.Background(),
+
+			releases:     releases,
+			oldVersion:   "12.1.1",
```
whites11, line 347: We allow skipping a patch when we are at the latest minor

fiunchinho (approved):
LGTM

internal/releaseversion/releaseversion.go
nprokopic: It is meant to be used for development and testing purposes and was never meant for production usage. We can maybe rename the annotation and add `unsupported` in the name.

whites11: restored

PR #141, opened 2021-01-15
Ensure autoscaler min and max annotations are present 

whites11:
This is to fix the bug in 13.1.0 release that could end up having MachinePool CRs without the autoscaler annotations set.

Still to be tested on the field, but please have a look

### DIFF
--- a/pkg/machinepool/autoscaling.go
+++ b/pkg/machinepool/autoscaling.go
@@ -18,8 +18,8 @@ func escapeJSONPatchString(input string) string {
 	return input
 }
 
-// setDefaultSpecValues checks if some optional field is not set, and sets
-// default values defined by upstream Cluster API.
+// ensureAutoscalingAnnotations ensures the custom annotations used to determine the min and max replicas for
+// the cluster autoscaler are set in the Machinepool CR.
 func ensureAutoscalingAnnotations(m mutator.Mutator, machinePool *capiexp.MachinePool) []mutator.PatchOperation {
 	var patches []mutator.PatchOperation
 
--- a/pkg/machinepool/autoscaling.go
+++ b/pkg/machinepool/autoscaling.go
@@ -24,7 +24,7 @@ func ensureAutoscalingAnnotations(m mutator.Mutator, machinePool *capiexp.Machin
 	var patches []mutator.PatchOperation
 
 	// The replicas field could not be set, we default to 1.
-	clusterReplicas := int32(1)
+	clusterReplicas := int32(defaultReplicas)
 	if machinePool.Spec.Replicas != nil {
 		clusterReplicas = *machinePool.Spec.Replicas
 	}
--- a/pkg/machinepool/autoscaling.go
+++ b/pkg/machinepool/autoscaling.go
@@ -2,10 +2,12 @@ package machinepool
 
 import (
 	"fmt"
-	"github.com/giantswarm/apiextensions/v3/pkg/annotation"
-	"github.com/giantswarm/azure-admission-controller/pkg/mutator"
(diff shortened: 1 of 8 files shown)


pkg/machinepool/autoscaling.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Ensure the needed escapes are in place. See https://tools.ietf.org/html/rfc6901#section-3 .
+func escapeJSONPatchString(input string) string {
+	input = strings.ReplaceAll(input, "~", "~0")
+	input = strings.ReplaceAll(input, "/", "~1")
+
+	return input
+}
+
+// setDefaultSpecValues checks if some optional field is not set, and sets
```
tuommaki, line 19: At least the comment function name is out-of-date, but maybe the body could be more specific as well? Alternatively just drop it.

pkg/machinepool/autoscaling.go
```diff
@@ -0,0 +1,41 @@
...
+
+	if machinePool.Annotations[annotation.NodePoolMinSize] == "" {
+		m.Log("level", "debug", "message", fmt.Sprintf("setting MachinePool Annotation %s to %d", annotation.NodePoolMinSize, clusterReplicas))
+		patches = append(patches, *mutator.PatchAdd(fmt.Sprintf("/metadata/annotations/%s", escapeJSONPatchString(annotation.NodePoolMinSize)), clusterReplicas))
+	}
+
+	if machinePool.Annotations[annotation.NodePoolMaxSize] == "" {
+		m.Log("level", "debug", "message", fmt.Sprintf("setting MachinePool Annotation %s to %d", annotation.NodePoolMaxSize, clusterReplicas))
+		patches = append(patches, *mutator.PatchAdd(fmt.Sprintf("/metadata/annotations/%s", escapeJSONPatchString(annotation.NodePoolMaxSize)), clusterReplicas))
+	}
```
tuommaki, line 38: You need to account also for situation where minimum has been set but maximum not; i.e. you don't want to set `min > max`. Possibly also auto-correct situation where `min < 1`.

pkg/machinepool/autoscaling.go
```diff
@@ -0,0 +1,41 @@
...
+	return input
+}
+
+// setDefaultSpecValues checks if some optional field is not set, and sets
+// default values defined by upstream Cluster API.
+func ensureAutoscalingAnnotations(m mutator.Mutator, machinePool *capiexp.MachinePool) []mutator.PatchOperation {
+	var patches []mutator.PatchOperation
+
+	// The replicas field could not be set, we default to 1.
+	clusterReplicas := int32(1)
```
nprokopic, line 25: The default `Spec.Replicas` is set here https://github.com/giantswarm/azure-admission-controller/blob/2740aedcf2dcad316fd8355b77469962b8c91f3b/pkg/machinepool/defaults.go#L28

(but that `Spec.Replicas` value might not be still set in the CR if this is the same webhook request, so this check is also needed)

We should promote a common constant and use it in both places, as it's a default value according to the CAPI spec.

whites11: done

pkg/machinepool/autoscaling.go
whites11: fixed, thanks

pkg/machinepool/autoscaling.go
whites11: makes sense, added more checks.

tuommaki (approved):
LGTM

whites11:
> LGTM

thanks, testing on gremlin before I merge and release

PR #147, opened 2021-01-19
Ensure azure-operator label when updating Cluster and AzureCluster

fiunchinho:
With these changes the admission controller would automatically set the right azure-operator component label to `Cluster` and `AzureCluster`. The right value for the label is taken from the `Release` used.


tuommaki (approved):
Besides minor nits in imports, this seems good to me.

PR #164, opened 2021-02-05
Add cluster mutate

whites11:
Added missing mutating webhook declaration in the helm chart.
Added automatic adjusting of the cluster-operator label according to the release version for `Cluster` and `AzureCluster`.


fiunchinho (approved):
Nice!

PR #173, opened 2021-02-08
Use upstream validation for resources

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/15570

This PR makes use of the upstream CAPZ validation for resources (defaulting will be done in a separate PR), with a few exceptions:
- `AzureMachine.Spec.SSHPublicKey` is required to not be empty upstream, while we require it to be empty.
- `AzureCluster.Metadata.Name` must not start with a number, but we have many existing clusters that do this.

### DIFF
--- a/internal/test/azurecluster/builder.go
+++ b/internal/test/azurecluster/builder.go
@@ -17,6 +17,7 @@ type BuilderOption func(azureCluster *capzv1alpha3.AzureCluster) *capzv1alpha3.A
 func Name(name string) BuilderOption {
 	return func(azureCluster *capzv1alpha3.AzureCluster) *capzv1alpha3.AzureCluster {
 		azureCluster.ObjectMeta.Name = name
+		azureCluster.Spec.ResourceGroup = name
 		azureCluster.Labels[capiv1alpha3.ClusterLabelName] = name
 		azureCluster.Labels[label.Cluster] = name
 		azureCluster.Spec.ControlPlaneEndpoint.Host = fmt.Sprintf("api.%s.k8s.test.westeurope.azure.gigantic.io", name)
@@ -72,6 +73,16 @@ func BuildAzureCluster(opts ...BuilderOption) *capzv1alpha3.AzureCluster {
 			ControlPlaneEndpoint: capiv1alpha3.APIEndpoint{
 				Host: fmt.Sprintf("api.%s.k8s.test.westeurope.azure.gigantic.io", clusterName),
 				Port: 443,
+			},
+			NetworkSpec: capzv1alpha3.NetworkSpec{
+				APIServerLB: capzv1alpha3.LoadBalancerSpec{
+					Name: "LB",
+					SKU:  "Standard",
+					Type: "Public",
+					FrontendIPs: []capzv1alpha3.FrontendIP{
+						{},
+					},
+				},
 			},
 		},
 	}
--- a/internal/test/azurecluster/builder.go
+++ b/internal/test/azurecluster/builder.go
@@ -80,7 +80,9 @@ func BuildAzureCluster(opts ...BuilderOption) *capzv1alpha3.AzureCluster {
 					SKU:  "Standard",
 					Type: "Public",
 					FrontendIPs: []capzv1alpha3.FrontendIP{
-						{},
+						{
+							Name: "LB",
+						},
 					},
 				},
 			},
--- a/internal/test/azurecluster/builder.go
+++ b/internal/test/azurecluster/builder.go
@@ -10,6 +10,7 @@ import (
 	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
 
 	"github.com/giantswarm/azure-admission-controller/internal/test"
+	"github.com/giantswarm/azure-admission-controller/pkg/key"
 )
 
 type BuilderOption func(azureCluster *capzv1alpha3.AzureCluster) *capzv1alpha3.AzureCluster
@@ -21,10 +22,10 @@ func Name(name string) BuilderOption {
 		azureCluster.Labels[label.Cluster] = name
 		azureCluster.Spec.ResourceGroup = name
 		azureCluster.Spec.ControlPlaneEndpoint.Host = fmt.Sprintf("api.%s.k8s.test.westeurope.azure.gigantic.io", name)
-		azureCluster.Spec.NetworkSpec.APIServerLB.Name = fmt.Sprintf("%s-%s-%s", name, "API", "PublicLoadBalancer")
+		azureCluster.Spec.NetworkSpec.APIServerLB.Name = key.APIServerLBName(name)
 		azureCluster.Spec.NetworkSpec.APIServerLB.FrontendIPs = []capzv1alpha3.FrontendIP{
 			{
-				Name: fmt.Sprintf("%s-%s-%s-%s", name, "API", "PublicLoadBalancer", "Frontend"),
+				Name: key.APIServerLBFrontendIPName(name),
 			},
 		}
 		return azureCluster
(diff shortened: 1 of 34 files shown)


pkg/azurecluster/error.go
```diff
@@ -66,3 +69,14 @@ var unexpectedLocationError = &microerror.Error{
 func IsUnexpectedLocationError(err error) bool {
 	return microerror.Cause(err) == unexpectedLocationError
 }
+
+// CAPZIsNameFieldValidationError asserts the CAPZ Name field validation error.
+func CAPZIsNameFieldValidationError(err error) bool {
+	if status := apierrors.APIStatus(nil); errors.As(err, &status) {
+		causes := status.Status().Details.Causes
+
+		return len(causes) > 0 && causes[0].Field == "metadata.Name"
```
axbarsan, line 78: 😅

pkg/azurecluster/validate_update_test.go
```diff
@@ -29,23 +29,11 @@ func TestAzureClusterUpdateValidate(t *testing.T) {
...
 			oldAzureCluster: builder.BuildAzureClusterAsJson(builder.Name("ab123"), builder.ControlPlaneEndpoint("api.ab123.k8s.test.westeurope.azure.gigantic.io", 443)),
 			newAzureCluster: builder.BuildAzureClusterAsJson(builder.Name("ab123"), builder.ControlPlaneEndpoint("api.ab123.k8s.test.westeurope.azure.gigantic.io", 80)),
 			errorMatcher:    IsControlPlaneEndpointWasChangedError,
 		},
-		{
-			name:            "case 3: location changed",
-			oldAzureCluster: builder.BuildAzureClusterAsJson(builder.Name("ab123"), builder.Location("westeurope")),
-			newAzureCluster: builder.BuildAzureClusterAsJson(builder.Name("ab123"), builder.Location("westpoland")),
-			errorMatcher:    IsLocationWasChangedError,
-		},
```
axbarsan, line 48: These 2 are already executed and tested in the upstream validation, no need to test them ourselves.

pkg/azurecluster/validate_create.go
```diff
@@ -58,7 +58,14 @@ func (a *CreateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
 		return microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureCluster CR: %v", err)
 	}
 
-	err := generic.ValidateOrganizationLabelContainsExistingOrganization(ctx, a.ctrlClient, azureClusterCR)
+	err := azureClusterCR.ValidateCreate()
+	if CAPZIsNameFieldValidationError(err) {
```
axbarsan, line 62: This validation error came up when using clusters that had uppercase characters in their names. Let's skip it for now.

pkg/azuremachine/validate_create.go
```diff
@@ -59,7 +59,14 @@ func (a *CreateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
 		return microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureMachine CR: %v", err)
 	}
 
-	err := generic.ValidateOrganizationLabelContainsExistingOrganization(ctx, a.ctrlClient, cr)
+	err := cr.ValidateCreate()
+	if CAPZIsSSHFieldValidationError(err) {
```
axbarsan, line 63: This validation error came up when using an empty SSH field, which is something we actually require to be empty. Let's ignore it for now.

internal/test/azurecluster/builder.go
```diff
@@ -73,6 +80,18 @@ func BuildAzureCluster(opts ...BuilderOption) *capzv1alpha3.AzureCluster {
...
+					Type: "Public",
+					FrontendIPs: []capzv1alpha3.FrontendIP{
+						{
+							Name: fmt.Sprintf("%s-%s-%s-%s", clusterName, "API", "PublicLoadBalancer", "Frontend"),
+						},
+					},
+				},
+			},
 		},
 	}
```
axbarsan, line 96: Taken from here https://github.com/giantswarm/azure-operator/blob/master/service/controller/azureconfig/handler/deployment/template/main.json#L1100-L1117

pkg/azurecluster/location.go
```diff
@@ -12,11 +12,3 @@ func validateLocation(azureCluster capzv1alpha3.AzureCluster, expectedLocation s
 
 	return nil
 }
-
-func validateLocationUnchanged(oldAzureCluster capzv1alpha3.AzureCluster, newAzureCluster capzv1alpha3.AzureCluster) error {
```
axbarsan, line 16: This is already done in the upstream validation

pkg/azurecluster/validate_create.go
fiunchinho: what if we don't ignore and add that validation? I don't think we use uppercase characters in cluster names anyway. What would that imply? Updating docs? WDYT?

axbarsan: Turns out our generated IDs don't have uppercase characters anyway (https://github.com/giantswarm/kubectl-gs/blob/2543fd3ba51cefb5494a1fdde43b148fdff334a2/internal/key/key.go#L29 and https://github.com/giantswarm/api/blob/8ff9eb0d8f25ebe65abe390eb7b0f8e38406425b/pkg/entityid/generator.go#L26). It was an error that only occured in tests. I'll update the code

internal/test/random.go
```diff
@@ -4,7 +4,7 @@ import (
 	"math/rand"
 )
 
-var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
fiunchinho, line 7: good catch

fiunchinho (changes requested):
You old friend the CHANGELOG needs a visit

pkg/azurecluster/validate_create.go
axbarsan: Ah wait, the name must not start or end with a number, which our generated IDs sometimes do. I need to update that ID generation

fiunchinho: Oh you are right about starting with a number, but I think it allos ending with a number.

axbarsan: It allows ending with a number, that's true. But we have a problem. All the existing clusters that have an ID that starts with a number will fail to validate if anyone tries to update them.

The error message says that it can begin with a alphanumeric character (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/417eccbe0ca8a433f44d1fb2a3a473c494a5ada3/api/v1alpha3/azurecluster_validation.go#L78), but that isn't true, since it can't begin with a number. 

Is the validation wrong, or the error message? 😅

fiunchinho: Let's ignore that error for now so our clusters won't fail, and bring this upstream. WDYT?

fiunchinho (approved):
LGTM

pkg/azuremachine/validate_update.go
```diff
@@ -52,7 +52,14 @@ func (a *UpdateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
 		return microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureMachine CR: %v", err)
 	}
 
-	err := checkSSHKeyIsEmpty(ctx, azureMachineNewCR)
+	err := azureMachineNewCR.ValidateUpdate(azureMachineOldCR)
+	if CAPZIsSSHFieldValidationError(err) {
```
tuommaki, line 56: Did you check if upstream validation executes all steps and returns collection of errors in the end or if the validation stops on first error?

Could it happen that due to "ignored error" rest of the CAPZ validation chain ends up being silently ignored?

tuommaki (approved):
No objections if all upstream validation steps are always executed. Good work! :+1:

pkg/azuremachine/validate_update.go
axbarsan: Damn, you're right... All validation errors are skipped if there is an SSH field error there. I'll fix it 😬

axbarsan: Please take a look

pkg/key/key.go
```diff
@@ -14,3 +20,47 @@ func GetControlPlaneEndpointHost(clusterName string, baseDomain string) string {
 func ServiceDomain() string {
 	return "cluster.local"
 }
+
+func IgnoreCAPIErrorForField(field string, err error) error {
```
axbarsan, line 24: God help us

fiunchinho: Can we move this function to `error.go`?

tuommaki: +1

tuommaki (approved):
LGTM, but please move the error matcher to `error.go` as suggested by Jose :wink:

pkg/azuremachine/mutate_update.go
```diff
@@ -0,0 +1,80 @@
...
+	ctrlClient ctrl.Client
+	logger     micrologger.Logger
+}
+
+type UpdateMutatorConfig struct {
+	CtrlClient ctrl.Client
+	Logger     micrologger.Logger
+}
+
+func NewUpdateMutator(config UpdateMutatorConfig) (*UpdateMutator, error) {
```
fiunchinho, line 26: I think you missed actually creating and wiring an instance of this mutator. Remember that it also needs to be added in the helm chart IIRC

axbarsan: Ah yes 😅

fiunchinho (approved):
Nice job man!

PR #178, opened 2021-02-15
Use CAPI/CAPZ defaulting for all resources

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/15570

This PR adds support for CAPI/CAPZ defaulting on all resources that support it. The patches are automatically generated based on the differences in the resources from before to after applying the defaults.

There are some messages below that hopefully explain my logic.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/giantswarm/micrologger v0.5.0
 	github.com/google/go-cmp v0.5.4
 	github.com/stretchr/testify v1.7.0
+	gomodules.xyz/jsonpatch/v2 v2.0.1
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	k8s.io/api v0.18.9
 	k8s.io/apiextensions-apiserver v0.18.9
--- a/helm/azure-admission-controller/templates/webhook.yaml
+++ b/helm/azure-admission-controller/templates/webhook.yaml
@@ -99,6 +99,24 @@ webhooks:
         - CREATE
   sideEffects: None
   admissionReviewVersions: ["v1", "v1beta1"]
+- name: mutate.azuremachinepools.update.{{ include "resource.default.name" . }}.giantswarm.io
+  failurePolicy: Fail
+  clientConfig:
+    service:
+      name: {{ include "resource.default.name" . }}
+      namespace: {{ include "resource.default.namespace" . }}
+      path: /mutate/azuremachinepool/update
+    caBundle: Cg==
+  rules:
+    - apiGroups: ["exp.infrastructure.cluster.x-k8s.io"]
+      resources:
+        - "azuremachinepools"
+      apiVersions:
+        - "v1alpha3"
+      operations:
+        - UPDATE
+  sideEffects: None
+  admissionReviewVersions: ["v1", "v1beta1"]
 - name: mutate.clusters.create.{{ include "resource.default.name" . }}.giantswarm.io
   failurePolicy: Fail
   clientConfig:
--- a/internal/patches/patches.go
+++ b/internal/patches/patches.go
@@ -1,0 +1,47 @@
+package patches
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/giantswarm/azure-admission-controller/pkg/mutator"
+	"github.com/giantswarm/microerror"
+	jsonpatch "gomodules.xyz/jsonpatch/v2"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func GeneratePatchesFrom(originalJSON []byte, current runtime.Object) ([]mutator.PatchOperation, error) {
+	currentJSON, err := json.Marshal(current)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	var patches []mutator.PatchOperation
+	{
+		var jsonPatches []jsonpatch.JsonPatchOperation
+		jsonPatches, err = jsonpatch.CreatePatch(originalJSON, currentJSON)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		patches = make([]mutator.PatchOperation, 0, len(jsonPatches))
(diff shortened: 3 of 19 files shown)


go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/giantswarm/micrologger v0.5.0
 	github.com/google/go-cmp v0.5.4
 	github.com/stretchr/testify v1.7.0
+	gomodules.xyz/jsonpatch/v2 v2.0.1
```
axbarsan, line 22: I haven't been able to automatically generate patches with our current `github.com/evanphx/json-patch` library.

Also, this new one is the same library [used by `controller-runtime` to generate patches](https://github.com/kubernetes-sigs/controller-runtime/blob/e1a725df2743147795e5dfc8275365f7ada24805/pkg/webhook/admission/response.go#L22).

helm/azure-admission-controller/templates/webhook.yaml
```diff
@@ -99,6 +99,24 @@ webhooks:
         - CREATE
   sideEffects: None
   admissionReviewVersions: ["v1", "v1beta1"]
+- name: mutate.azuremachinepools.update.{{ include "resource.default.name" . }}.giantswarm.io
```
axbarsan, line 102: Adding a new `UPDATE` webhook for `AzureMachinePool` resources.

pkg/azurecluster/mutate_create.go
```diff
@@ -139,6 +140,20 @@ func (m *CreateMutator) Mutate(ctx context.Context, request *v1beta1.AdmissionRe
...
+	azureClusterCR.Default()
+	{
+		var capiPatches []mutator.PatchOperation
+		capiPatches, err = patches.GeneratePatchesFrom(request.Object.Raw, azureClusterCR)
+		if err != nil {
+			return []mutator.PatchOperation{}, microerror.Mask(err)
+		}
+
+		capiPatches = patches.SkipPatchesForPath("/spec/networkSpec/vnet", capiPatches)
+		capiPatches = patches.SkipPatchesForPath("/spec/networkSpec/subnets", capiPatches)
```
axbarsan, line 152: Ignoring these, as we set their configuration inside `azure-operator`, and we already default to a `control-plane` subnet in the code above.

pkg/azurecluster/mutate_update.go
```diff
@@ -111,6 +112,20 @@ func (m *UpdateMutator) Mutate(ctx context.Context, request *v1beta1.AdmissionRe
...
+	azureClusterCR.Default()
+	{
+		var capiPatches []mutator.PatchOperation
+		capiPatches, err = patches.GeneratePatchesFrom(request.Object.Raw, azureClusterCR)
+		if err != nil {
+			return []mutator.PatchOperation{}, microerror.Mask(err)
+		}
+
+		capiPatches = patches.SkipPatchesForPath("/spec/networkSpec/vnet", capiPatches)
+		capiPatches = patches.SkipPatchesForPath("/spec/networkSpec/subnets", capiPatches)
```
axbarsan, line 124: Same here

pkg/azuremachine/mutate_create.go
```diff
@@ -83,6 +84,19 @@ func (m *CreateMutator) Mutate(ctx context.Context, request *v1beta1.AdmissionRe
...
 
+	azureMachineCR.Default()
+	{
+		var capiPatches []mutator.PatchOperation
+		capiPatches, err = patches.GeneratePatchesFrom(request.Object.Raw, azureMachineCR)
+		if err != nil {
+			return []mutator.PatchOperation{}, microerror.Mask(err)
+		}
+
+		capiPatches = patches.SkipPatchesForPath("/spec/sshPublicKey", capiPatches)
```
axbarsan, line 95: We require users to not set this value. We have validation to make sure it's empty, while CAPZ defaults it to a generic key.

pkg/machinepool/defaults.go
```diff
@@ -15,21 +13,5 @@ const defaultReplicas int64 = 1
 func setDefaultSpecValues(m mutator.Mutator, machinePool *capiexp.MachinePool) []mutator.PatchOperation {
 	var patches []mutator.PatchOperation
 
-	defaultSpecReplicas := setDefaultReplicaValue(m, machinePool)
```
axbarsan, line 18: This is already set in the CAPI defaulting logic.

internal/patches/patches.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		patches = make([]mutator.PatchOperation, 0, len(jsonPatches))
+		for _, patch := range jsonPatches {
+			patches = append(patches, mutator.PatchOperation(patch))
+		}
+
+		sort.SliceStable(patches, func(i, j int) bool {
```
axbarsan, line 34: Without sorting, the order of the patches varies from one test run to another, which causes the tests to sometimes fail.

internal/patches/patches.go
```diff
@@ -0,0 +1,53 @@
...
+	"strings"
+
+	"github.com/giantswarm/microerror"
+	jsonpatch "gomodules.xyz/jsonpatch/v2"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/giantswarm/azure-admission-controller/pkg/mutator"
+)
+
+func GeneratePatchesFrom(originalJSON []byte, current runtime.Object) ([]mutator.PatchOperation, error) {
```
tuommaki, line 15: Minor nitpick: When package is `patches`, it might make sense to drop it from the function name:
`patches.GeneratePatchesFrom()` -> `patches.GenerateFrom()`

Same below for `SkipPatchesForPath`

See: https://blog.golang.org/package-names#TOC_3.

internal/patches/patches.go
tuommaki: Good catch :+1:

tuommaki (approved):

axbarsan:
Tested again ✅

PR #179, opened 2021-02-16
Don't validate the 'spec.networkSpec.subnets' field of 'AzureCluster' resources

axbarsan:
For solving the problem found here https://gigantic.slack.com/archives/CNUDS3AMU/p1613459439057600

Currently, subnets are added to the `AzureCluster` CRs by `AzureMachinePool` CRs via `azure-operator`, which goes against the validation logic in here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Execute CAPZ validation for all resources.
+- Execute CAPI/CAPZ validation for all resources.
+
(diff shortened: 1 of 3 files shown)


whites11 (approved):
Guess it works for now

tuommaki (approved):
I think dropping this for now is a good thing, but we might want to add an issue to our board to later revisit if this could be re-activated again?

axbarsan:
> I think dropping this for now is a good thing, but we might want to add an issue to our board to later revisit if this could be re-activated again?

I created this https://github.com/giantswarm/giantswarm/issues/15758 for that 👍🏻

PR #182, opened 2021-02-16
Skip CAPI/CAPZ validation for the 'AzureCluster.Spec.SubscriptionID' field

axbarsan:
Solves the problem found here https://gigantic.slack.com/archives/CNUDS3AMU/p1613475275079500

This field is actually set by the `azure-operator` after taming some monsters and doing some dark magic.


tuommaki (approved):
:hugs:

PR #190, opened 2021-02-23
Just default the whole 'azureCluster.spec.networkSpec.apiServerLB' field

axbarsan:
As seen here https://gigantic.slack.com/archives/C03CPNRTS/p1614073534058200

Tested and it works ✅


whites11 (approved):
LGTMIIWFY

PR #198, opened 2021-03-09
Update architect-orb to 2.1.0

ubergesundheit:
Update architect-orb to 2.1.0 which will use giantswarm.github.io as catalog base url. The old behaviour was to use giantswarm.github.com which will be deprecated in april by GitHub. Towards giantswarm/giantswarm#15898

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@0.17.0
+  architect: giantswarm/architect@2.1.0
 
 
 workflows:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -48,7 +48,6 @@ workflows:
           context: architect
           app_name: "azure-admission-controller"
           app_collection_repo: "azure-app-collection"
-          unique: "true"
```
fiunchinho, line 51: we don't need to specify this anymore?

ubergesundheit: nope, this was removed in architect-orb 1.0.0 https://github.com/giantswarm/architect-orb/blob/master/CHANGELOG.md#removed-1

PR #210, opened 2021-04-08
Ignore CAPI/CAPZ CRs with watch filter label

nprokopic:
When creating CAPI+CAPZ clusters, AzureMachine CRs that are created by MachineSet by using AzureMachineTemplate do not have release label, so azure-admission-controller is blocking CR creation due to validation errors.

This PR adds following checks:
- It checks for `cluster.x-k8s.io/watch-filter` label which is added by Kyverno defaulting, so that azure-admission-controller will ignore newly created CRs that are part of new CAPI+CAPZ clusters. Just checking release label is not enough, since it is not added to the CRs during defaulting.
- Edit: Adding above check was not enough because apparently at the moment `cluster.x-k8s.io/watch-filter` is not set on AzureMachine when the creation request gets to the azure-admission-controller. "Incorrect" ordering of admission controller calls may be an issue here, if azure-admission-controller is called before Kyverno. So I added another check for `cluster.x-k8s.io/cloned-from-name` annotation, which is set by CAPI controller when an AzureMachine is created from an AzureMachineTemplate. We are not setting this annotation, so using it here for checking if a CR belongs to CAPI release should be fine.

The PR also includes architect bump which then required a bunch of code refactoring to make new linter happy :slightly_smiling_face:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Skipping all validation and defaulting for resources with CAPI release label.
 - Skipping all validation and defaulting for resources with CAPI watch filter label.
-- Update architect version to 2.7.0.
+- Refactor code to make new linter happy.
 
 ## [2.3.2] - 2021-02-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Skipping all validation and defaulting for resources with CAPI release label.
-- Skipping all validation and defaulting for resources with CAPI watch filter label.
+- Skipping all validation and defaulting for resources with the release label containing a release `>=` than the first GiantSwarm release with CAPI controllers.
+- Skipping all validation and defaulting for resources with the CAPI watch filter label.
 - Refactor code to make new linter happy.
 
 ## [2.3.2] - 2021-02-23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Skipping all validation and defaulting for resources with CAPI release label
-- Skipping all validation and defaulting for resources with CAPI watch filter label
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Skipping all validation and defaulting for resources with CAPI release label
+- Skipping all validation and defaulting for resources with CAPI release label.
+- Skipping all validation and defaulting for resources with CAPI watch filter label.
+- Update architect version to 2.7.0.
```
whites11, line 14: This is a CI bump and should't be in the CHANGELOG imho

whites11 (approved):

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Skipping all validation and defaulting for resources with CAPI release label
+- Skipping all validation and defaulting for resources with CAPI release label.
+- Skipping all validation and defaulting for resources with CAPI watch filter label.
```
fiunchinho, line 13:
```suggestion
- Skipping all validation and defaulting for resources with the release label containing a release `>=` than the first GiantSwarm release with CAPI controllers.
- Skipping all validation and defaulting for resources with the CAPI watch filter label.
```

nprokopic: I just added a dot on the end of that sentence :smile:

anvddriesch:
LGTM! Just double check that it works with the order in which the webhooks are called. (kyverno adds the label and only after that azure-admission-controller can check it)

CHANGELOG.md
fiunchinho: I know, I know, but after your changes it wasn't clear to me what the first sentence was trying to explain.

fiunchinho (approved):

tuommaki (approved):

nprokopic:
@giantswarm/team-celestial please take a look if `cluster.x-k8s.io/cloned-from-name` check is fine here. I could not think of any other workaround to figure out if the CR is from a CAPI release.

tuommaki (approved):
Seems fine for me

nprokopic:
Closing this in favor of new PR https://github.com/giantswarm/azure-admission-controller/pull/216.

PR #211, opened 2021-04-08
Allow empty datadisks for AzureMachinePool

tuommaki:


pkg/azuremachinepool/datadisks.go
```diff
@@ -24,6 +24,10 @@ var desiredDataDisks = []capzv1alpha3.DataDisk{
 }
 
 func checkDataDisks(ctx context.Context, mp *expcapzv1alpha3.AzureMachinePool) error {
+	if len(mp.Spec.Template.DataDisks) == 0 {
```
whites11, line 27: this doesn't feel right to me. The following `if` checks if the value of the field matches a desired value (that I assume is not empty). I feel like this additional check is changing the logic

nprokopic:
I'm working on these CircleCI issues in https://github.com/giantswarm/azure-admission-controller/pull/210

> ERROR: The key you are authenticating with has been marked as read only.
> fatal: Could not read from remote repository.

pkg/azuremachinepool/datadisks.go
fiunchinho: Yeah we created this to make sure `AzureMachinePool` always have the desired data disks. Should we only apply the former rule to objects containing older-than-CAPI releases?

tuommaki: > The following if checks if the value of the field matches a desired value

True, but the message says explicitly `AzureMachinePool.Spec.Template.DataDisks is reserved and should not be set.` and when I check that the field is not set, I'd assume that to be valid.

Let's fix the message then :slightly_smiling_face: 

> Should we only apply the former rule to objects containing older-than-CAPI releases?

We probably need to refactor the admission controller and provide some way to control which rules apply to which releases, but first we need to have the discussion on if we will continue with this or Kyverno.

fiunchinho: It seems we've come to an agreement so KaaS teams can use their own admission controllers, so we can keep using this Golang based controller.

PR #216, opened 2021-04-20
CR filtering & spring cleaning

nprokopic:
Closes https://github.com/giantswarm/giantswarm/issues/16844

## 1. Overview

With this change `azure-admission-controller` will process only CRs that belong to a cluster from a release that contains `azure-operator`.

The release to which the CR belongs to will be determined in one of the following ways, whichever is possible:
1. check `release.giantswarm.io/version` label from the CR itself
2. get `cluster.x-k8s.io/cluster-name` from the CR, and then get the Cluster CR by name and check its `release.giantswarm.io/version` label
3. get `giantswarm.io/cluster` from the CR, and then get the Cluster CR by name and check its `release.giantswarm.io/version` label

If release cannot be determined in one of the aforementioned 3 ways, `azure-admission-controller` will skip the CR.

Hard requirement for this logic to work is that Cluster CR has `release.giantswarm.io/version` label set. Here we have two paths:
1. `release.giantswarm.io/version` label is set during Cluster CR creation - AFAIK this is currently how our tooling works.
2. If not set on Cluster CR creation, `release.giantswarm.io/version` label is defaulted by Kyverno and it is set to the latest release version. In this case we would probably need to set webhook `reinvocationPolicy` to `IfNeeded`, so that `azure-admission-controller` is invoked again after mutating Cluster CR to set `release.giantswarm.io/version`.

## 2. Implementation details

Some parts of the azure-admission-controller had a bit bigger changes, like changing function signature in mutators and validators `Mutate` and `Validate` functions, moving some checks and conversions from individual validator/mutator implementation one level above to the mutation/validation handler func, so it's implemented in a centralized place, instead of repeating the same thing many times.

### 2.1. Validators & mutators interfaces

`validator.Validator` and `mutator.Mutator` interfaces have been renamed to `validator.WebhookHandler` and `mutator.WebhookHandler` and have been refactored from this:

```
type Validator interface {
	Validate(ctx context.Context, request *v1beta1.AdmissionRequest) error
	Log(keyVals ...interface{})
}

type Mutator interface {
	Log(keyVals ...interface{})
	Mutate(ctx context.Context, review *v1beta1.AdmissionRequest) ([]PatchOperation, error)
	Resource() string
}
```

to this:
```
type Decoder interface {
	Decode(rawObject runtime.RawExtension) (metav1.ObjectMetaAccessor, error)
}

type Logger interface {
	Log(keyVals ...interface{})
}

type WebhookHandler interface {
	generic.Decoder
	generic.Logger
}

type WebhookCreateHandler interface {
	WebhookHandler
	OnCreateValidate(ctx context.Context, object interface{}) error
}

type WebhookUpdateHandler interface {
	WebhookHandler
	OnUpdateValidate(ctx context.Context, oldObject interface{}, object interface{}) error
}

type WebhookHandler interface {
	generic.Decoder
	generic.Logger
	Resource() string
}

type WebhookCreateHandler interface {
	WebhookHandler
	OnCreateMutate(ctx context.Context, object interface{}) ([]PatchOperation, error)
}

type WebhookUpdateHandler interface {
	WebhookHandler
	OnUpdateMutate(ctx context.Context, oldObject interface{}, object interface{}) ([]PatchOperation, error)
}
```

You can find these new interfaces here:
- logger and decoder https://github.com/giantswarm/azure-admission-controller/blob/fdbd8893d34daf324915b5ad3b6a91337cf5c9f4/pkg/generic/types.go
- mutation webhook handler https://github.com/giantswarm/azure-admission-controller/blob/fdbd8893d34daf324915b5ad3b6a91337cf5c9f4/pkg/mutator/webhook_handler.go
- validation webhook handler https://github.com/giantswarm/azure-admission-controller/blob/fdbd8893d34daf324915b5ad3b6a91337cf5c9f4/pkg/validator/webhook_handler.go

Every CR type has one struct that is implementing both validator and mutator interfaces for both create and update operations (only those that are required), so that there is less repeating code around new types.

All of this was done to facilitate the changes in the validator/mutator HTTP handlers, which you will see in the next two sections.

### 2.2. Validators & mutators implementation

For every CR type, we used to have 2 validator implementations, one for create and one for update, and 2 mutator implementation, one for create and one for update. All of these implementation were almost identical, with same fields in structures. For every type, these 4 different structs has been merged into a single one.

For example, instead of having:
- `cluster.CreateMutator` with `Mutate(...)` func
- `cluster.UpdateMutator` with `Mutate(...)` func
- `cluster.CreateValidator` with `Validate(...)` func
- `cluster.UpdateValidator` with `Validate(...)` func

Now there is a single `cluster.WebhookHandler` type with these 4 functions:
- `OnCreateMutate(...)`
- `OnUpdateMutate(...)`
- `OnCreateValidate(...)`
- `OnUpdateValidate(...)`

These 4 functions are coming from 4 different interfaces, so `sometype.WebhookHandler` can only implement those that are needed, which preserves the flexibility.

### 2.3. CR deserialization/decoding

Every validator and mutator implementation has a code like this:
```
azureClusterCR := &capzv1alpha3.AzureCluster{}
if _, _, err := validator.Deserializer.Decode(rawObject.Raw, nil, azureClusterCR); err != nil {
	return nil, microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureCluster CR: %v", err)
}
```

This has been now moved to a separate function that implements `Decoder` interface:
```
func (h *WebhookHandler) Decode(rawObject runtime.RawExtension) (metav1.ObjectMetaAccessor, error) {
	azureClusterCR := &capz.AzureCluster{}
	if _, _, err := validator.Deserializer.Decode(rawObject.Raw, nil, azureClusterCR); err != nil {
		return nil, microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureCluster CR: %v", err)
	}

	return azureClusterCR, nil
}
```

With this change CR decoding is called in a validation/mutation HTTP handler. More about "why" in the next section.

### 2.4. CR filtering

As state previously, validator/mutator HTTP handler is calling CR decoding. **This enables the HTTP handler to do CR filtering in one place, instead of every validator/mutator checking the CR on its own** - the HTTP handler is checking if the CR should be processed (validated or mutated) at all, here's a snippet:
```
object, err := webhookCreateHandler.Decode(review.Request.Object)
if err != nil {
    return microerror.Mask(err)
}

ok, err := filter.IsCRProcessed(ctx, h.ctrlClient, object.GetObjectMeta())
if err != nil {
    return microerror.Mask(err)
}

if ok {
    err = webhookCreateHandler.OnCreateValidate(ctx, object)
    if err != nil {
        return microerror.Mask(err)
    }
}
```

This way the HTTP handler is directly taking care of the CR decoding and filtering, and the individual validator/mutator implementation are called only when the CR should be processed, and they get already decoded CRs, so they only need to assert them to specific types, similarly to what we do in operators based on operatorkit.
```
func (a *Validator) OnCreateValidate(ctx context.Context, object interface{}) error {
	clusterCR, err := key.ToClusterPtr(object)
	if err != nil {
		return microerror.Mask(err)
	}
// ...
```

Filtering implementation can be found here https://github.com/giantswarm/azure-admission-controller/blob/fdbd8893d34daf324915b5ad3b6a91337cf5c9f4/pkg/filter/filter.go.

### 2.5. Mutating dry run

Mutating dry run check has been moved from individual mutator implementations one level up to the common mutator handler, as the behavior is the same for all types.

### 2.6. Moving functions around

Some functions has been moved to different packages, e.g. some functions for ensuring labels have been moved from `pkg/generic` to `pkg/mutator`, as I got into a situation to have some circular dependencies and those functions are specific to and used in mutators anyway.

### 2.7. Minor test improvements

The tests logic is not modified, all tests should be identical as they were. One minor change that has been made is how the test `WebhookHandler` (previously `Mutator` and `Validator`) objects have been constructed before calling tested function.

We used to have something like this in many places:
```
validator := &UpdateValidator{
    logger: newLogger,
    ctrlClient: fakeCtrlClient,
    // someField which is not set
    // someOtherField which is not set
}
```

So we would construct the minimum object that we can so we can test `validator.Validate(...)` function without panics. This assumes knowledge about internal implementation details and relying on the fact that some of the fields are not used in the function that is being tested.

This seems little wrong and fragile, as the unit test should not care about internal implementation details and it should test the function like a black box, by providing needed input and checking its output. For this we need a fully constructed object on which we will call the function that we want to test.

These cases have been changed to look like more like this (what we already do in many places as well):
```
handler, err := NewWebhookHandler(WebhookHandlerConfig{
    BaseDomain: "k8s.test.westeurope.azure.gigantic.io",
    CtrlClient: ctrlClient,
    Logger:     newLogger,
})
if err != nil {
    t.Fatal(err)
}

// Run admission request to validate Cluster updates.
err = handler.OnCreateValidate(ctx, &tc.cluster)
```

So basically constructing the object with provided public API (`NewWebhookHandler` here) and calling the function we want to test.


pkg/filter/filter.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
+	capiutil "sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-admission-controller/pkg/release"
+)
+
+func IsCRProcessed(ctx context.Context, ctrlClient client.Client, objectMeta metav1.Object) (bool, error) {
```
nprokopic, line 16: CR filtering implementation:
- it first checks for release label
- if previous not found, then it checks GS cluster label (so it can check Cluster's release label)
- if previous not found, then it checks CAPI cluster label (so it can check Cluster's release label)

pkg/filter/filter.go
```diff
@@ -0,0 +1,55 @@
...
+		if releaseVersionLabel == "" {
+			// We found the cluster, but we cannot find out which release the
+			// cluster belongs to.
+			return false, nil
+		}
+	}
+
+	// Now when we have release version for the CR, let's check if the release
+	// contains azure-operator.
+	return release.ContainsAzureOperator(ctx, ctrlClient, releaseVersionLabel)
```
nprokopic, line 54: After getting the release, it checks if the release contains `azure-operator`.

pkg/release/release.go
```diff
@@ -4,35 +4,70 @@ import (
...
+	gocache "github.com/patrickmn/go-cache"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var (
+	// Cache of release components
+	// key: release version
+	// value: map "component name": "component version"
+	releaseComponentsCache = gocache.New(24*time.Hour, 24*time.Hour)
```
nprokopic, line 20: Release components are cached, as they are checked in every request now.

pkg/validator/handler.go
```diff
@@ -26,7 +37,81 @@ var (
...
+	}
+
+	h := &HandlerFactory{
+		ctrlClient: config.CtrlClient,
+	}
+
+	return h, nil
+}
+
+func (h *HandlerFactory) NewCreateHandler(validator CreateValidator) http.HandlerFunc {
```
nprokopic, line 60: Separate `NewCreateHandler` and `NewUpdateHandler` functions, because due to the second one having to decode one more object, and they call different funcs - `Validate` and `ValidateUpdate` respectively.

pkg/mutator/handler.go
```diff
@@ -15,22 +15,114 @@ import (
...
+	}
+
+	h := &HandlerFactory{
+		ctrlClient: config.CtrlClient,
+	}
+
+	return h, nil
+}
+
+func (h *HandlerFactory) NewCreateHandler(mutator CreateMutator) http.HandlerFunc {
```
nprokopic, line 67: Separate `NewCreateHandler` and `NewUpdateHandler` functions, because due to the second one having to decode one more object, and they call different funcs - `Mutate` and `MutateUpdate` respectively.

nprokopic:
I'll test it more, especially "live" on test installations, but it's ready for review @giantswarm/team-celestial.

main.go
```diff
@@ -129,27 +129,15 @@ func mainError() error {
 		}
 	}
 
-	var azureClusterCreateMutator *azurecluster.CreateMutator
+	var azureClusterMutator *azurecluster.Mutator
```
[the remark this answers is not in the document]
tuommaki: I tried to quickly search but failed: Where are these interfaces defined? Currently `Mutate()` and `MutateUpdate()` feel a bit too generic and ambiguous. Could the wording be slightly improved:
When type is `Mutator` and interfaces `CreateMutator` and `UpdateMutator` (if I got it right), how about following:

```go
type CreateMutator interface {
    OnCreate(ctx context.Context, object interface{}) ([]PatchOperation, error)
}

type UpdateMutator interface {
    OnUpdate(ctx context.Context, oldObject interface{}, object interface{}) ([]PatchOperation, error)
}
```

The code later would be then `mutator.OnCreate(ctx, cr)` or `mutator.OnUpdate(ctx, cr)`.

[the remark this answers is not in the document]
nprokopic: Here are the interfaces https://github.com/giantswarm/azure-admission-controller/blob/fcd84278adb2ab88d5315d1627481e070585ed46/pkg/mutator/handler.go#L24-L38

> Currently Mutate() and MutateUpdate() feel a bit too generic and ambiguous.

It's actually less generic now, it was just `Mutator` with `Mutate` func, I broke it down a bit more :smile: 

I like the `OnCreate` and `OnUpdate` suggestions.

WDYT about also renaming `Mutator` to `MutatingWebhookHandler`, and `Validator` to `ValidatingWebhookHandler`?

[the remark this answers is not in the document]
tuommaki: > WDYT about also renaming Mutator to MutatingWebhookHandler, and Validator to ValidatingWebhookHandler?

Works for me :+1:

pkg/key/key.go
```diff
@@ -40,3 +51,107 @@ func OSDiskCachingType() string {
 func MasterSubnetName(clusterName string) string {
 	return fmt.Sprintf("%s-%s-%s", clusterName, "VirtualNetwork", "MasterSubnet")
 }
+
+func ToClusterPtr(v interface{}) (*capi.Cluster, error) {
```
nprokopic, line 55: Just doing type assertion and returning pointers.

pkg/machinepool/mutate_machinepool_update.go
```diff
@@ -1,91 +0,0 @@
-package machinepool
```
nprokopic, line 1: This is not deleted, but it's renamed to `mutate_update.go`.

pkg/machinepool/validate_machinepool_create.go
```diff
@@ -1,119 +0,0 @@
-package machinepool
```
nprokopic, line 1: This is not deleted, bit it's renamed to `validate_create.go`.

pkg/mutator/http_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	h := &HttpHandlerFactory{
+		ctrlClient: config.CtrlClient,
+	}
+
+	return h, nil
+}
+
+func (h *HttpHandlerFactory) NewHttpCreateHandler(mutator WebhookCreateHandler) http.HandlerFunc {
```
nprokopic, line 50: Creating HTTP handler for mutating create requests.

pkg/mutator/http_handler.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+		}
+
+		return patch, nil
+	}
+
+	return h.newHttpHandler(mutator, mutateFunc)
+}
+
+func (h *HttpHandlerFactory) NewHttpUpdateHandler(mutator WebhookUpdateHandler) http.HandlerFunc {
```
nprokopic, line 77: Creating HTTP handler for mutating update requests (different than create as it also needs old CR).

pkg/release/error.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 var releaseNotFoundError = &microerror.Error{
-	Kind: "downgradingIsNotAllowedError",
+	Kind: "releaseNotFoundError",
```
nprokopic, line 8: :broom:

pkg/validator/http_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	h := &HttpHandlerFactory{
+		ctrlClient: config.CtrlClient,
+	}
+
+	return h, nil
+}
+
+func (h *HttpHandlerFactory) NewCreateHttpHandler(webhookCreateHandler WebhookCreateHandler) http.HandlerFunc {
```
nprokopic, line 47: Creating HTTP handler for validating create requests.

pkg/validator/http_handler.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+		}
+
+		return nil
+	}
+
+	return h.newHttpHandler(webhookCreateHandler, validateFunc)
+}
+
+func (h *HttpHandlerFactory) NewUpdateHttpHandler(webhookUpdateHandler WebhookUpdateHandler) http.HandlerFunc {
```
nprokopic, line 72: Creating HTTP handler for validating update requests (different than create as it also needs old CR).

pkg/validator/http_handler.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (h *HttpHandlerFactory) NewCreateHttpHandler(webhookCreateHandler WebhookCreateHandler) http.HandlerFunc {
+	validateFunc := func(ctx context.Context, review v1beta1.AdmissionReview) error {
+		object, err := webhookCreateHandler.Decode(review.Request.Object)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		ok, err := filter.IsCRProcessed(ctx, h.ctrlClient, object.GetObjectMeta())
```
nprokopic, line 54: Checking if `azure-admission-controller` should validate CR. If this returns `false`, validation is not called at all.

nprokopic: Same is done for update below.

pkg/mutator/http_handler.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func (h *HttpHandlerFactory) NewHttpCreateHandler(mutator WebhookCreateHandler) http.HandlerFunc {
+	mutateFunc := func(ctx context.Context, review v1beta1.AdmissionReview) ([]PatchOperation, error) {
+		object, err := mutator.Decode(review.Request.Object)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		ok, err := filter.IsCRProcessed(ctx, h.ctrlClient, object.GetObjectMeta())
```
nprokopic, line 57: Checking if `azure-admission-controller` should mutate CR. If this returns `false`, mutation is not called at all. Same is done for update below.

tuommaki (approved):
I only skimmed through this due to size. It could help to distinct individual pattern-like changes into separate PRs so that it's easy to quickly read through each of them as only one thing changes at a time.

Maybe like: 
- One PR to change `capiv1alpha3` -> `capi` & `capzv1alpha3` -> `capz`.
- One PR for introducing new webhook type (without usage).
- Distinct PRs to introduce new webhook functions for existing handlers, one function per PR.
  - e.g. `Decode()` for all handlers.
- One PR to refactor unit tests if there's a need?
  - The tests that work at handler top level shouldn't need much changing.
- One PR to refactor usage of old interfaces to new one.
- Last PR to drop any left over old code?

nprokopic:
> I only skimmed through this due to size. It could help to distinct individual pattern-like changes into separate PRs so that it's easy to quickly read through each of them as only one thing changes at a time.

Yeah, it would be a lot easier to review this if it was split into multiple PRs. The hard part of it is that most of the changes are breaking, so it's hard to have a PR with only one set of individual pattern-like changes like you suggest, e.g. changing former validators and mutators to webhool handlers breaks `main.go` and even tests for that validator and mutator.

In any case the splitting would be very good, I'll see what I can do. Splitting per CRD could be easier maybe *that's how I implemented it, CRD by CRD).

> Maybe like:
> 
>     * One PR to change `capiv1alpha3` -> `capi` & `capzv1alpha3` -> `capz`.
> 
>     * One PR for introducing new webhook type (without usage).
> 
>     * Distinct PRs to introduce new webhook functions for existing handlers, one function per PR.
>       
>       * e.g. `Decode()` for all handlers.
> 
>     * One PR to refactor unit tests if there's a need?
>       
>       * The tests that work at handler top level shouldn't need much changing.
> 
>     * One PR to refactor usage of old interfaces to new one.
> 
>     * Last PR to drop any left over old code?

These suggestions could be possible probably, although there could be a lot of double/copied if we want multiple PRs where every one works on its own.

PR #218, opened 2021-05-07
Don't validate CRs when they are being deleted

fiunchinho:
When developing, sometimes it's a pain in the ass to delete resources because the admission controller gets in the way (i.e the `Release` CR no longer exists).
So let's disable validation when deleting some CR's.

### DIFF
--- a/pkg/azurecluster/validate_update.go
+++ b/pkg/azurecluster/validate_update.go
@@ -53,6 +53,11 @@ func (a *UpdateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
 		return microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureCluster CR: %v", err)
 	}
 
+	if !azureClusterNewCR.GetDeletionTimestamp().IsZero() {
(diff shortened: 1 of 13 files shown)


whites11 (approved):
<3

pkg/azurecluster/validate_update.go
```diff
@@ -53,6 +53,11 @@ func (a *UpdateValidator) Validate(ctx context.Context, request *v1beta1.Admissi
 		return microerror.Maskf(errors.ParsingFailedError, "unable to parse AzureCluster CR: %v", err)
 	}
 
+	if !azureClusterNewCR.GetDeletionTimestamp().IsZero() {
```
fiunchinho, line 56: I'm not sure if it would be better to check both `azureClusterOldCR` and `azureClusterNewCR` , but the only situation where `azureClusterNewCR` would have `DeletionTimestamp` set and `azureClusterOldCR` wouldn't, would be when deleting the CR. With current code, in that case it would be validated one last time.

Thoughts?

whites11: Yeah if we are deleting it we don't care about it being valid any more

tuommaki (approved):

PR #225, opened 2021-05-14
Use new config system

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/16669

Can't be merged until this is merged https://github.com/giantswarm/config/pull/88


whites11 (approved):
LGTM

helm/azure-admission-controller/values.yaml
```diff
@@ -1,7 +1,25 @@
...
+  service:
+    azure:
+      clientid: ""
+      clientsecret: ""
+      sptenantid: ""
+      subscriptionid: ""
+      tenantid: ""
+
+cluster:
+  baseDomain: ""
```
tuommaki, line 22: Minor nitpick: Is there something behind style difference here: `clientsecret` vs. `baseDomain` - why not `clientSecret` or `basedomain` ?

tuommaki (approved):

helm/azure-admission-controller/values.yaml
fiunchinho: Not really. The reason is that I was using camelCase but I also copied the `clientid`, `clientsecret`, etc from `azure-operator`, and they are defined without camelCase over there. I'll change them to `clientID`, `clientSecret`, etc.

PR #246, opened 2021-05-21
Add WebhookHandler types

nprokopic:

## Motivation

https://github.com/giantswarm/azure-admission-controller/pull/216 introduces a breaking change where all mutators and validators are merged into a single `WebhookHandler` per CRD. That breaking change then propagates through the entire `azure-admission-controller`, it requires updates in tests and main for example.

This PR only introduces new `WebhookHandler` interfaces for the sake of easier reviewing, without actually using them anywhere.

## What's in the box

- `validator.WebhookCreateHandler` interface for validating `create` requests
- `validator.WebhookUpdateHandler` interface for validating `update` requests
- `mutator.WebhookCreateHandler` interface for mutating `create` requests
- `mutator.WebhookUpdateHandler` interfaces for mutating `update` requests
- `generic.Decoder` interface to abstract away CR decoding from the admission request
- `Logger` cause we log a lot :)

The interfaces are intentionally split so that there is one for every validate/mutate + verb combination. This way we can only implement webhooks that we really need (YAGNI :slightly_smiling_face:).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When importing `sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3` use `capzexp` as package alias.
 - Remove package names from some file names.
 - Move labels mutator functions to `pkg/mutator`.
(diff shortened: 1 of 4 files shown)


tuommaki (approved):
Nice :+1:

PR #247, opened 2021-05-21
Add CR filtering functionality

nprokopic:

## Motivation

We want that `azure-admission-controller` only validates and mutates CRs that belong to clusters from "legacy" non-CAPI releases, because CRs from new CAPI cluters will be validated and mutated by upstream webhooks, our new admission controller, and Kyverno.

## What's in the box?

- New `pkg/filter` package with `IsObjectReconciledByLegacyRelease (context.Context, client.Client, metav1.Object) (bool, error)` function which checks if the CR belongs to a cluster from a legacy non-CAPI release. The function is still not used anywhere.
  - Release is considered to be "legacy" if it contains azure-operator.
- Caching release components for 24 hours, as this will be used in all admission controller requests.


nprokopic:
I would like to add few more tests around release component caching, but feel free to review this already.

pkg/filter/filter.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/giantswarm/micrologger"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
+	capiutil "sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-admission-controller/pkg/release"
+)
+
+func IsCRProcessed(ctx context.Context, ctrlClient client.Client, logger micrologger.Logger, objectMeta metav1.Object) (bool, error) {
```
tuommaki, line 17:
```suggestion
func IsObjectReconciledByLegacyRelease(ctx context.Context, ctrlClient client.Client, logger micrologger.Logger, objectMeta metav1.Object) (bool, error) {
```

Minor nitpick: `IsCRProcessed` makes me think this function checks if the admission controller already processed given `object` while this is more about checking if the object is reconciled by legacy release. WDYT?

pkg/release/release.go
```diff
@@ -4,14 +4,26 @@ import (
...
 )
 
-func GetComponentVersionsFromRelease(ctx context.Context, ctrlClient client.Client, releaseVersion string) (map[string]string, error) {
+var (
+	cacheExpiry = 24 * time.Hour
+
+	// Cache of release components
+	// key: release version
+	// value: map "component name": "component version"
+	releaseComponentsCache = gocache.New(cacheExpiry, 24*time.Hour)
```
tuommaki, line 23: Is there a mechanism that refreshes the cached release if it changes? I'm afraid this caching is a dangerous footgun during development.

Could there be a simple watcher running in the background instead, that would persist & refresh present `Releases` in an in-memory map, which would then be served like the API below?

nprokopic: > Is there a mechanism that refreshes the cached release if it changes?

AFAIK there isn't (I did not implement anything in azure-admission-controller and I don't think there is anything like that in the cache library).

> I'm afraid this caching is a dangerous footgun during development.
I can't say I understand why :confused:

I believe that when the cache item expires the `cache.Get(key)` will not return any result. When that happens `azure-admission-controller` will just call API server again and cache the results again for the next 24h, here:

https://github.com/giantswarm/azure-admission-controller/pull/247/files/b53c0880f1affe47aaf7d45950332baed6d85302#diff-c1b95cb221c72b791b38018615bb6bb835b5f303ff0cd75871cbb1f8edbb8514R67-R81

Maybe there is a better cache expiry time than 24h, this was just a semi-wild semi-educated guess :)

tuommaki: Sorry, I should have been more elaborate:

When the release is cached for some period of time, it affects the behavior and defaulting of admission controller here. For most of the time it works fine and we forget about it. Then at some point we need to do some bisecting or frequent updating of component for a given release and we observe weird behavior. If we don't remember that there exists this cache inside admission controller (and that it requires re-start of the whole program in order to flush it), we can easily waste a lot of time looking elsewhere for the root cause of this weird behavior.

If the implementation had watch + map instead of on-demand cache, the releases would be effectively cached, but also always up-to-date.

I'm not strictly requiring this, but just saying that there is a potential here for overlooking once we forget the existence of this cache :slightly_smiling_face:

pkg/filter/filter.go
nprokopic: :heavy_check_mark:

pkg/release/release.go
nprokopic: Oh I see now, so while we are working on a component `X`, frequently changing its version in a release during develpment and its version in the release is cached in the azure-admission-controller, weird stuff might happen because that version will not be refreshed in the cache. Makes sense.

What kind of watch implementation would you suggest?

I just wouldn't like to complicate it to the point where it's just better not to do caching at all, because its benefits would be almost lost due to increased complexity of the implementation which adds another type of burden and issues :smile: OTOH simple but flaky behavior (like you described) is also not desireable.

pkg/release/release.go
```diff
@@ -36,3 +36,21 @@ func GetComponentVersionsFromRelease(ctx context.Context, ctrlReader client.Read
 
 	return ret, nil
 }
+
+// ContainsAzureOperator checks if the specified release contains azure-operator.
+//
+// In order to perform the check, this function is calling GetComponentVersionsFromRelease function,
+// which is caching obtained components in memory. See GetComponentVersionsFromRelease docs for
+// more info about the caching.
```
tuommaki, line 44: Is this comment valid when the caching actually happens in the passed client which can be either caching one or the "normal" one (same interface)?

pkg/filter/filter.go
```diff
@@ -0,0 +1,55 @@
...
+		if releaseVersionLabel == "" {
+			// We found the cluster, but we cannot find out which release the
+			// cluster belongs to.
+			return false, nil
+		}
+	}
+
+	// Now when we have release version for the CR, let's check if the release
+	// contains azure-operator.
+	return release.ContainsAzureOperator(ctx, ctrlClient, releaseVersionLabel)
```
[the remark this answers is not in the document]
nprokopic: Yes, I overlooked it when I rebased my PR. Thanks!

pkg/release/release.go
nprokopic: True, this is not needed anymore.

pkg/filter/filter.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
+	capiutil "sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-admission-controller/pkg/release"
+)
+
+func IsObjectReconciledByLegacyRelease(ctx context.Context, ctrlClient client.Client, ctrlCache client.Reader, objectMeta metav1.Object) (bool, error) {
```
nprokopic, line 16: @tuommaki  I added both `client.Client` and `client.Reader` here since CAPI `GetClusterByName(...)` func needs `client.Client` and we need `client.Reader` for cached releases.

Not sure if there is some better way to handle this.

pkg/filter/filter_test.go
```diff
@@ -0,0 +1,422 @@
...
+			loadReleases(t, ctrlClient)
+
+			if tc.ownerCluster != nil {
+				err := ctrlClient.Create(ctx, tc.ownerCluster)
+				if err != nil {
+					t.Fatalf("Error when setting up test data: %#v", err)
+				}
+			}
+
+			result, err := IsObjectReconciledByLegacyRelease(ctx, ctrlClient, ctrlClient, tc.inputCR)
```
nprokopic, line 255: @tuommaki I passed in `client.Client` here for the reader as well, since here we are testing the filtering logic and not the caching logic.

pkg/filter/filter.go
nprokopic: Wait, I'll just not use CAPI helper func, since it's a regular Get call inside, that will make the func signature prettier.

tuommaki: What if this was split into two functions: `FindClusterRelease()` and `IsLegacyRelease()` ?

nprokopic: We can, just not sure if there is a need, as this is the CAPI helper func

```
// GetClusterByName finds and return a Cluster object using the specified params.
func GetClusterByName(ctx context.Context, c client.Client, namespace, name string) (*clusterv1.Cluster, error) {
	cluster := &clusterv1.Cluster{}
	key := client.ObjectKey{
		Namespace: namespace,
		Name:      name,
	}

	if err := c.Get(ctx, key, cluster); err != nil {
		return nil, err
	}

	return cluster, nil
}
```

It's literally just a Get call :smile:

nprokopic: I can split the func here and make the consumer call 2 funcs, but I think it would be more clear to have one func that answers the question `Should I validate/mutate this or just ignore it?`

pkg/release/release.go
```diff
@@ -6,11 +6,69 @@ import (
...
+// Cluster API controllers.
+func IsLegacyRelease(ctx context.Context, ctrlCache client.Reader, releaseVersion string) (bool, error) {
+	releaseContainsAzureOperator, err := ContainsAzureOperator(ctx, ctrlCache, releaseVersion)
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	return releaseContainsAzureOperator, nil
+}
+
```
nprokopic, line 71: @tuommaki Now I've split the filtering func into these two, as you suggested.

pkg/filter/filter.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/giantswarm/azure-admission-controller/pkg/release"
+)
+
+// IsObjectReconciledByLegacyRelease checks if the object is reconciled by an operator which is the
+// part of a legacy Giant Swarm release (a release that does not have Cluster API controllers).
+func IsObjectReconciledByLegacyRelease(ctx context.Context, ctrlCache client.Reader, objectMeta metav1.Object) (bool, error) {
```
nprokopic, line 15: @tuommaki the filtering func in now just calling two new funcs. Good suggestion to split this, it's easier to understand it this way.

tuommaki: The code is indeed more readable, but I think the original problem of two clients was slightly missed here. Now there's only one client used here and it's the caching one. This has the effect that now `Cluster` is also cached in the informer. It might not be a bad idea on itself, but should be conscious decision.

I'm thinking that these problems with two different clients and challenges in wiring / carrying them through call chains are a bit of a smell of non-optimal architecture. The functions we have here are theoretically just something that should work on plain data. One should be able to implement them in pure functions given that the input data was fetched beforehand.

The API of function signatures is as generic as it can be while somewhat typed, but the implementation itself is quite specific since it involves knowledge about `Cluster` type and how to find it.

Could the design be such that all necessary objects are fetched on top level and then passed as data to these functions? That way we could better control the API calls and avoid accidental re-fetching in different execution paths while making functions signatures more clear on intent and allowing easier testing of them.

Example:
Top level function to fetch the release. Note that `Cluster` was already fetched and the passed reader is only used to fetch the `Release` object itself.
`func FindRelease(ctx context.Context, ctrlCache client.Reader, cluster capi.Cluster, objectMeta metav1.Object) (releasev1alpha1.Release, error)`

The logic that knows whether it's legacy release or not. No other objects nor clients needed since the `Release` itself is here:
`func IsLegacyRelease(release releasev1alpha1.Release) (bool, error)`

I think it could be useful if we paired on this a bit so that there's not that long latency through asynchronous reviews?

nprokopic: > This has the effect that now Cluster is also cached in the informer. It might not be a bad idea on itself, but should be conscious decision.

Oh, I was not aware of that.

> One should be able to implement them in pure functions given that the input data was fetched beforehand.
> ...
> Could the design be such that all necessary objects are fetched on top level and then passed as data to these functions? That way we could better control the API calls and avoid accidental re-fetching in different execution paths while making functions signatures more clear on intent and allowing easier testing of them.

I see, yeah that would be probably better, to basically make the function a stateless "utility" function, and make consumers (that know about cached vs not cached) talk to the API.

tuommaki (approved):
LGTM :+1:

PR #257, opened 2021-05-31
Introduce watch based `release.Getter`

tuommaki:
This change introduces `release.Getter` interface that can be used to
abstract the method of `Release` acquisition.

To provide up-to-date caching implementation for frequent release
utilization on admission controller requests, the default implementation
for release getter is built on `controller-runtime` primitives. It
updates cached object on reconciliation event and deletes it from cache
when corresponding CR also receives non-zero `deletionTimestamp`.


tuommaki:
I think we need to learn a few things here and there. We ain't gonna need this. Closing here. :see_no_evil:

PR #258, opened 2021-05-31
Add simple `release.Getter` interface w/ caching client

tuommaki:
Utilize caching client for frequently used Release objects to avoid
unnecessary round-trips to apiserver while still keeping the objects
fresh in the cache.


pkg/release/caching_getter.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewCachingGetter(cfg *rest.Config) (Getter, error) {
+	scheme := runtime.NewScheme()
+	err := releasev1alpha1.AddToScheme(scheme)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	opts := cache.Options{
+		Scheme: scheme,
```
tuommaki, line 28: Unsure if I need to declare `RESTMapper` as well..

pkg/release/spec.go
```diff
@@ -0,0 +1,7 @@
+package release
+
+import releasev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/release/v1alpha1"
+
+type Getter interface {
```
tuommaki, line 5: My idea here is to provide only this simple interface for object getting, wire its implementation up to all components that need the cached client and finally separate the functions that utilize `Release` objects themselves.

pkg/release/caching_getter.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	c, err := cache.New(cfg, opts)
```
tuommaki, line 31: After I wrote the watch-based client, I had to re-check few implementation details and then I found this, which is exactly what I was working on, but more general. I ditched my original implementation and used this.

tuommaki:
Let's discuss about this in sync. I'm leaning towards dropping this one as well :smile:

PR #268, opened 2021-06-21
Add HTTP handler factory

nprokopic:

## Motivation

In order to use CR filtering from https://github.com/giantswarm/azure-admission-controller/pull/247 we need to refactor how we are creating HTTP handlers that are handling azure-admission-controller requests.

## What's in the box?

- `validator.HttpHandlerFactory` that creates HTTP handlers for validating create and update requests.
- `mutator.HttpHandlerFactory` that creates HTTP handlers for mutating create and update requests.

The logic how the HTTP handler is processing the requests is largely the same, with the addition of first decoding the CR and checking if the azure-admission-controller should validate/mutate it.


pkg/generic/meta.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type OwnerClusterGetter func(metav1.ObjectMetaAccessor) (capi.Cluster, bool, error)
+
+func TryGetClusterName(object metav1.ObjectMetaAccessor) (string, bool) {
```
nprokopic, line 16: I'll add code docs here

pkg/generic/meta.go
```diff
@@ -0,0 +1,60 @@
...
+		if clusterID != "" {
+			// We found GS cluster ID, this is our cluster name.
+			clusterName = clusterID
+		}
+	}
+
+	return clusterName, clusterName != ""
+}
+
+func GetOwnerCluster(ctx context.Context, ctrlClient client.Client, object metav1.ObjectMetaAccessor) (capi.Cluster, bool, error) {
```
nprokopic, line 37: and here

pkg/mutator/http_handler_factory.go
```diff
@@ -0,0 +1,183 @@
...
+			ownerCluster, ok, err := generic.GetOwnerCluster(ctx, h.ctrlClient, object)
+			if err != nil {
+				return capi.Cluster{}, false, microerror.Mask(err)
+			}
+
+			return ownerCluster, ok, nil
+		}
+
+		// Check if the CR should be mutated by the azure-admission-controller.
+		ok, err := filter.IsObjectReconciledByLegacyRelease(ctx, h.ctrlClient, object, ownerClusterGetter)
```
nprokopic, line 60: Hmm this is incorrect, cache should be passed here. Same in few other places. I'll fix that.

pkg/generic/meta.go
tuommaki, line 37: I'd maybe use `client.Reader` here instead of `client.Client` so it would be possible to use caching client if desired.

pkg/mutator/handler.go
```diff
@@ -97,7 +99,7 @@ func extractName(request *v1beta1.AdmissionRequest) string {
 	return "<unknown>"
 }
 
-func writeResponse(mutator Mutator, writer http.ResponseWriter, response *v1beta1.AdmissionResponse) {
+func writeResponse(mutator generic.Logger, writer http.ResponseWriter, response *v1beta1.AdmissionResponse) {
```
tuommaki, line 102: Variable name `mutator` _looks_ weird for a logger :stuck_out_tongue:

pkg/validator/handler.go
```diff
@@ -61,7 +63,7 @@ func Handler(validator Validator) http.HandlerFunc {
 	}
 }
 
-func writeResponse(validator Validator, writer http.ResponseWriter, response *v1beta1.AdmissionResponse) {
+func writeResponse(validator generic.Logger, writer http.ResponseWriter, response *v1beta1.AdmissionResponse) {
```
tuommaki, line 66: Similar to `mutator`, maybe rename `validator` to `logger` or something along the lines :slightly_smiling_face:

tuommaki (approved):
Couple minor nitpicks but so far LGTM

pkg/mutator/handler.go
nprokopic: true that :smile:

nprokopic: leftovers I think

pkg/generic/meta.go
fiunchinho: Since cluster name can't be empty, what about only returning `string` and make consumers check `if string != ""` instead of checking the returned bool value?

nprokopic: This felt more explicit, like trying to get env var with `os.LookupEnv("var_name")

pkg/mutator/testdata/release-v14.1.4.yaml
```diff
@@ -0,0 +1,58 @@
...
+    version: 2605.12.0
+  - name: calico
+    version: 3.15.3
+  - name: etcd
+    version: 3.4.14
+  date: "2021-04-06T08:44:15Z"
+  state: active
+status:
+  inUse: false
+  ready: false
```
fiunchinho, line 58: Today, in nitpicking news: [Why should text files end with a newline?](https://stackoverflow.com/a/729795/563072)

fiunchinho (approved):
LGTM

pkg/mutator/testdata/release-v14.1.4.yaml
nprokopic: That was a nice read there :grin:

PR #272, opened 2021-07-05
Add cluster.WebhookHandler

nprokopic:
Towards https://github.com/giantswarm/giantswarm/issues/17957

## Motivation

Here https://github.com/giantswarm/azure-admission-controller/pull/268 we have introduced a new HTTP handler factory that is creating new handlers that are using new `WebhookHandler` interfaces and which are making use of CR filtering functionality that is introduced here https://github.com/giantswarm/azure-admission-controller/pull/247.

Now it's time to put all of that into action and actually start using it in azure-admission-controller.

P.S. We are doing all of this so that azure-admission-controller is only validating and mutating the CRs that belong to legacy Giant Swarm releases. New CRs from new CAPI releases will be validated and mutated by new admission controllers.

## Implementation

This pull request adds implementation of these interfaces for the `Cluster` CRD:
- `mutator.WebhookCreateHandler`
- `mutator.WebhookUpdateHandler`
- `validator.WebhookCreateHandler`
- `validator.WebhookUpdateHandler`

The implementation of these interfaces will replace current `CreateMutator`, `UpdateMutator`, `CreateValidator` and `UpdateValidator`. The important part here is that the mutation and validation logic remains exactly the same, it's just refactored significantly in order to move to new HTTP handlers with CR filtering.

## Testing

Every `WebhookHandler` implementation for every CRD will be done in a separate pull request. All these pull requests will be stacked on top of each other, so that every new is added on top of the previous one. This way it will be possible to test all of them together (which is needed, because it does not make sense to do the filtering on only some CRDs), but implement and review the changes separately.

We also have to adapt existing unit and integration tests.

Therefore we have the following testing tasks:
- [x] update all existing unit tests for `Cluster` to use new `WebhookHandler` implementations
- [x] update all existing and affected integration tests to use new `WebhookHandler` implementations
- [x] test changes in test installations

Important note:
*This pull request will be merged only when the testing of all the following related pull requests for all other CRDs is completed (every PR builds on top of the previous one):
- Cluster _(this PR)_
- AzureCluster https://github.com/giantswarm/azure-admission-controller/pull/273
- MachinePool https://github.com/giantswarm/azure-admission-controller/pull/274
- AzureMachinePool https://github.com/giantswarm/azure-admission-controller/pull/275
- AzureMachine https://github.com/giantswarm/azure-admission-controller/pull/278
- Spark https://github.com/giantswarm/azure-admission-controller/pull/279
- AzureClusterConfig https://github.com/giantswarm/azure-admission-controller/pull/280
- AzureConfig https://github.com/giantswarm/azure-admission-controller/pull/281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Release is considered to be "legacy" if it contains azure-operator.
 - Unit tests for functions from `release` package.
 - `HttpHandlerFactory` for creating HTTP handlers that are using new webhook handlers.
+- Cluster webhook handler that replaces mutators and validators.
 
 ### Changed
 
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -148,3 +148,12 @@ func IgnoreCAPIErrorForField(field string, err error) error {
 
 	return err
 }
+
+var WrongTypeError = &microerror.Error{
+	Kind: "WrongTypeError",
+}
+
+// IsWrongTypeError asserts wrongTypeError.
+func IsWrongTypeError(err error) bool {
+	return microerror.Cause(err) == WrongTypeError
+}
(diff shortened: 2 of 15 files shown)


fiunchinho (approved):
This looks great!

tuommaki (approved):
:rocket:

PR #282, opened 2021-07-16
Refactor main and unit tests

nprokopic:
## Motivation

It happens every now and then that integration tests fail in CircleCI because some object did not have all of its dependencies wired in. Since code in `main.go` is not tested, those issues could not discovered in unit tests.

Another issue that happened in the past is that some webhook handler (previously mutator and/or validator) would not be registered at all, so although the code is there, it was not being executed.

## Implementation

This PR changes the following:
- All webhook handlers (previously mutators and validators) are created in `pkg/app` package. There is a unit test that checks if this works, so we can now see failures more quickly, and more easily, e.g. with `make test` on our dev machines.
- The webhook handlers are not added to HTTP handler manually. Now for all handlers we check which wehbook handler interfaces are implementing, and according to that we add appropriate HTTP handlers.
- Resource SKU stub API has been moved to `pkg/unittests` since it was being used (and copy-pasted) in multiple places.


pkg/app/handlers.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/giantswarm/azure-admission-controller/pkg/azureupdate"
+	"github.com/giantswarm/azure-admission-controller/pkg/cluster"
+	"github.com/giantswarm/azure-admission-controller/pkg/config"
+	"github.com/giantswarm/azure-admission-controller/pkg/machinepool"
+	"github.com/giantswarm/azure-admission-controller/pkg/mutator"
+	"github.com/giantswarm/azure-admission-controller/pkg/spark"
+	"github.com/giantswarm/azure-admission-controller/pkg/validator"
+)
+
+// RegisterWebhookHandlers first creates all required webhook handlers and then it adds registers
```
fiunchinho, line 25:
```suggestion
// RegisterWebhookHandlers first creates all required webhook handlers and then it registers
```

pkg/app/handlers_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Fatal(microerror.JSON(err))
+	}
+
+	// Real *http.ServeMux, not that we gonna run it here.
+	handler := http.NewServeMux()
+
+	// Run webhook handlers registration.
+	err = RegisterWebhookHandlers(handler, cfg, logger, ctrlClient, ctrlClient, vmcaps)
+	if err != nil {
+		t.Errorf("Error while registering webhook handlers %#v", err)
```
fiunchinho, line 48:
```suggestion
		t.Fatalf("Error while registering webhook handlers %#v", err)
```

pkg/azurecluster/mutate_create_test.go
```diff
@@ -133,7 +133,7 @@ func TestAzureClusterCreateMutate(t *testing.T) {
 
 			handler, err := NewWebhookHandler(WebhookHandlerConfig{
 				BaseDomain: "k8s.test.westeurope.azure.gigantic.io",
-				CtrlCache:  ctrlClient,
+				CtrlReader: ctrlClient,
```
fiunchinho, line 136: why is it sometimes called `CtrlReader` and sometimes `CtrlCache`?

nprokopic: Yeah, I noticed the same, so I changed these to `CtrlReader` everywhere, or at least I thought I did :smile:

nprokopic:
@giantswarm/team-celestial any other issue here?

fiunchinho (approved):

PR #286, opened 2021-07-26
Testing CRs in real management clusters

nprokopic:
## Motivation

It is possible that some change in azure-admission-controller may cause some undesired changes and issues in CR filtering, or validating and mutating webhook handlers, which may affect only a portion of CRs in a management cluster.

Issues like this are hard to catch in regular unit tests.

## Solution (to some extent)

The goal of this PR is to add tests that can catch if some issue would happen on a real management cluster. It does that with tests that are using real CRs from real management clusters as test cases.

- [x] CR filtering tests for all Cluster, AzureCluster, MachinePool, AureMachinePool, AzureMachine, Spark and AzureConfig CRs in all management clusters
- [x] WebhookHandler tests for all Cluster, AzureCluster, MachinePool, AureMachinePool, AzureMachine, Spark and AzureConfig CRs in all management clusters

For the sake of not modifying these objects by accident, the tests are using internal `client.Client` implementation that allows only read-only operations (e.g. `client.Get(...)` and `client.List(...)`) and it causes the test to fail if any modification operation is called (e.g. `client.Create(...)`, `client.Update(...)`, `client.Delete(...)`, etc..

AzureClusterConfig validation is not tested, since it is not needed and it will be removed.

Make command to run the tests:

```
make test-live-crs MANAGEMENT_CLUSTER="mymanagementclustername"
```

You need `opsctl` installed to run `make test-live-crs`.

## Testing

- [x] Run tests for all CRs in all management clusters

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,14 @@ include Makefile.*.mk
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-.PHONY: live-installation-test
-## live-installation-test: runs CR validation tests on a live installation
+.PHONY: test-live-crs
+## test-live-crs: runs CR validation tests on a real management cluster
 test-live-crs:
 	@echo "====> $@"
-	go test -ldflags "$(LDFLAGS)" -tags="liveinstallation" -race ./integration/test/validateliveresources
+ifeq ($(MANAGEMENT_CLUSTER),)
+	@echo "MANAGEMENT_CLUSTER not set, exiting"
+else
+	@echo "Testing CRs in ${MANAGEMENT_CLUSTER}"
+	opsctl create kubeconfig -i "${MANAGEMENT_CLUSTER}"
+	go test -count=1 -ldflags "$(LDFLAGS)" -tags="liveinstallation" -race ./integration/test/validateliveresources
+endif
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,9 @@ help: ## Display this help.
 test-live-crs:
 	@echo "====> $@"
 ifeq ($(MANAGEMENT_CLUSTER),)
-	@echo "MANAGEMENT_CLUSTER not set, exiting"
+	@echo "MANAGEMENT_CLUSTER not set, using current kubectl context."
 else
-	@echo "Testing CRs in ${MANAGEMENT_CLUSTER}"
+	@echo "Testing CRs in ${MANAGEMENT_CLUSTER}, changing kubectl context."
 	opsctl create kubeconfig -i "${MANAGEMENT_CLUSTER}"
-	go test -count=1 -ldflags "$(LDFLAGS)" -tags="liveinstallation" -race ./integration/test/validateliveresources
 endif
+	go test -count=1 -ldflags "$(LDFLAGS)" -tags="liveinstallation" -race ./integration/test/validateliveresources
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,6 @@ help: ## Display this help.
 
 .PHONY: live-installation-test
 ## live-installation-test: runs CR validation tests on a live installation
-live-installation-test:
+test-live-crs:
 	@echo "====> $@"
 	go test -ldflags "$(LDFLAGS)" -tags="liveinstallation" -race ./integration/test/validateliveresources
--- a/Makefile
+++ b/Makefile
@@ -21,3 +21,9 @@ include Makefile.*.mk
 .PHONY: help
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+.PHONY: live-installation-test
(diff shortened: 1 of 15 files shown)


integration/test/validateliveresources/azureconfig_test.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, azureConfig := range azureConfigList.Items {
+		if !azureConfig.GetDeletionTimestamp().IsZero() {
+			// Skip CRs that are being deleted.
+			continue
+		}
+
+		updatedAzureConfig := azureConfig.DeepCopy()
+		updatedAzureConfig.Labels["test.giantswarm.io/dummy"] = "this is not really saved"
+
+		// Test mutating webhook, on create.
```
fiunchinho, line 98: I think it's not mutating nor creating :smile:

integration/test/validateliveresources/azurecluster_test.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		result, err := filter.IsObjectReconciledByLegacyRelease(ctx, logger, ctrlClient, &azureCluster, ownerClusterGetter)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if result == false {
+			objectName := fmt.Sprintf("%s/%s", azureCluster.Namespace, azureCluster.Name)
+			t.Errorf("Expected 'true' (AzureCluster %s is reconciled by legacy release), got 'false' "+
```
fiunchinho, line 56: I believe you are trying to bee too explicit and it becomes even confusing. I think `Expected AzureCluster %s to be reconciled by legacy release` tells what happened, without having to expose the true/false that IMO is not needed when reading test failures.

integration/test/validateliveresources/helpers.go
```diff
@@ -0,0 +1,161 @@
...
+	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
+	capiexp "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+
+	"github.com/giantswarm/azure-admission-controller/integration/env"
+	"github.com/giantswarm/azure-admission-controller/internal/vmcapabilities"
+)
+
+func NewReadOnlyCtrlClient(t *testing.T) client.Client {
```
fiunchinho, line 33: Could we use the controller-runtime reader/cache thingy instead? That interface didn't have the writing methods if I remember correctly? So we wouldn't need our own implementation

fiunchinho (commented):
I don't know how I feel about this. It feels weird to access a real management cluster. A real MC is a live thing which changes all the time so I'm not sure we'd be covering that much cases anyway. But I'm not opposing to keep it.

Makefile
```diff
@@ -21,3 +21,15 @@ include Makefile.*.mk
...
+## test-live-crs: runs CR validation tests on a real management cluster
+test-live-crs:
+	@echo "====> $@"
+ifeq ($(MANAGEMENT_CLUSTER),)
+	@echo "MANAGEMENT_CLUSTER not set, exiting"
+else
+	@echo "Testing CRs in ${MANAGEMENT_CLUSTER}"
+	opsctl create kubeconfig -i "${MANAGEMENT_CLUSTER}"
+	go test -count=1 -ldflags "$(LDFLAGS)" -tags="liveinstallation" -race ./integration/test/validateliveresources
+endif
```
fiunchinho, line 35: Instead of requiring the management cluster to use, could we use the current context instead? I don't really like `opsctl`, changing the current context.

nprokopic: I was doing that all the time while I was testing, and since it became a bit annoying I figured one-line make command with less previous required steps would be useful.

Let's meet half way and do what you say, use current context, when `MANAGEMENT_CLUSTER` var is not specified, how does that sound?

integration/test/validateliveresources/helpers.go
nprokopic: > That interface didn't have the writing methods if I remember correctly? So we wouldn't need our own implementation

True, I started with that one, then saw that we use `client.Client` in all structs in the project, so it's not possible to pass `client.Reader`. We would have to change all existing `client.Client` usage to `client.Reader` instead, which would probably be even better, as we would be sure that we are not modifying stuff accidentally.

Another caveat there is that we also already have `client.Reader` in some structs, which we use for cache purposes (since `client.Reader` is a subset interface of `client.Cache`).
So if we don't want to have two `client.Reader` fields in some places, we would have to also change `client.Reader` into `client.Cache` there, aaand this was the part where I stopped, since I was not fully sure why Tuomas used `client.Reader` instead of `client.Cache` in the first place, but one thing that I'm sure is that the tests are simpler with `client.Reader`.

So there you go :slightly_smiling_face: I took a bit shorter path (this little read-only wrapper struct) as you can see :smile:

integration/test/validateliveresources/azurecluster_test.go
nprokopic: Good point, I was changing this text few times and I got a Frankenstein in the end :grimacing:

Makefile
nprokopic: Better now?

nprokopic:
> I don't know how I feel about this. It feels weird to access a real management cluster. A real MC is a live thing which changes all the time so I'm not sure we'd be covering that much cases anyway. But I'm not opposing to keep it.

It happened few times in the past that we have break something in azure-admission-controller and catch it only when we actually deploy it. In those cases validation and/or mutation for already existing CRs would stop working (maybe even without any change in those CRs). Tests like these help to catch those issues.
Think of it like a convenient CLI helper (in a form of `make` command) that you can just run and see what happens, instead of deploying entire azure-admission-controller and checking logs and etc. Additional benefit is that you can run this on a production cluster as well, since it is read-only. So we can catch possible issues before deploying it.

> A real MC is a live thing which changes all the time so I'm not sure we'd be covering that much cases anyway

IMO exactly because the CRs (and the components that use/change them) are changing all the time we cannot predict every test case in our regular unit/e2e test cases. So something like this, on top of our existing unit/e2e tests, can help us to cover more possible cases by testing real production CRs.

fiunchinho (approved):
Ok, LGTM

PR #287, opened 2021-07-29
Migrate to config management

corest:
Towards https://github.com/giantswarm/giantswarm/issues/18230

How can I fix integration test? I tested it on real env and didn't have issues


corest:
Integration test issue was related to wrong values

PR #321, opened 2021-10-04
Added validation for update-schedule annotation

whites11:
This PR adds a validation on the scheduled upgrade annotations.


whites11:
Integration tests fail because it can't download `kind` and I have no clue why

internal/scheduledupgrades/scheduledupgrades.go
```diff
@@ -0,0 +1,94 @@
...
+	// parse time
+	t, err := time.Parse(time.RFC822, updateTime)
+	if err != nil {
+		return false
+	}
+	// check whether it is UTC
+	if t.Location().String() != "UTC" {
+		return false
+	}
+	//time already passed or is less than 16 minutes in the future
```
fiunchinho, line 50:
```suggestion
	// time already passed or is less than 16 minutes in the future
```

internal/scheduledupgrades/scheduledupgrades.go
```diff
@@ -0,0 +1,94 @@
...
+			return microerror.Maskf(notAllowedError,
+				fmt.Sprintf("Cluster annotation '%s' value '%s' is not valid. Value must be an existing giant swarm release version above the current release version %s and must not have a v prefix. %v",
+					annotation.UpdateScheduleTargetRelease,
+					targetRelease,
+					oldVersion.String(),
+					err),
+			)
+		}
+
+		if oldVersion.Equals(newVersion) {
```
fiunchinho, line 84: isn't this already covered in `releaseversion.Validate()`?

internal/scheduledupgrades/scheduledupgrades.go
```diff
@@ -0,0 +1,94 @@
...
+		if oldCluster != nil {
+			if updateTimeOld, ok := oldCluster.GetAnnotations()[annotation.UpdateScheduleTargetTime]; ok {
+				if updateTime == updateTimeOld {
+					return nil
+				}
+			}
+		}
+		if !ValidateUpgradeScheduleTime(updateTime) {
+			return microerror.Maskf(notAllowedError,
+				fmt.Sprintf("Cluster annotation '%s' value '%s' is not valid. Value must be in RFC822 format and UTC time zone (e.g. 30 Jan 21 15:04 UTC) and should be a date 16 mins - 6months in the future.",
```
fiunchinho, line 31: where do these rules come from? have we documented it somewhere?

internal/scheduledupgrades/scheduledupgrades.go
whites11: it's the opposite check. Within releaseversion.Validate() we accept the fact that release is unchanged, but in this case we want release to be changed or there is no upgrade to be done.

internal/scheduledupgrades/scheduledupgrades.go
whites11: This method call comes straight from AWS admission controller. No idea about documentation

fiunchinho (approved):

PR #332, opened 2021-11-16
Add more replicas and introduce pdb

njuettner:


helm/azure-admission-controller/templates/deployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "labels.common" . | nindent 4 }}
 spec:
-  replicas: 1
+  replicas: 3
```
whites11, line 9: Why 3 ? Isn't 2 enough with this PDB?

njuettner: That was the consent because of the amount of masters running in production, we can also 2 replicas it should be enough 👍🏻.

njuettner, line 9:
```suggestion
  replicas: 2
```

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add pod disruption budget.
+
+### Changed
+
+- Adjust the number of replicas to 3.
```
njuettner, line 16:
```suggestion
- Adjust the number of replicas to 2.
```

whites11 (approved):
Thank you Nick

PR #336, opened 2021-11-24
Split VM type capabilities client by subscription.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/19932

On Azure, availability zones change their name on a per subscription basis.
In azure-admission-controller we used to have one single set of azure credentials (the MC credentials) in order to check if a certain VM type was available on a certain availability zone.
This is wrong, as with BYOC organizations the subscription may be different and lead to legit requests being blocked or invalid requests being allowed.

This PR replaces the global "VM Capabilites" helper (previously initialized in the main function and propagated to all webhook handlers) with a factory that allows each handler to initialize an helper instance every time it is needed, using the specific credentials for the subscription.

The PR is huge (sorry :pray:) but most of the changed files are just wiring changes and the interesting files are just a handful.


internal/capzcredentials/capz.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/giantswarm/microerror"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capz "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetAzureCredentialsFromMetadata(ctx context.Context, ctrlClient client.Client, obj metav1.ObjectMeta) (string, string, string, string, error) {
```
njuettner, line 14: what about putting the 4 strings into a struct so you can return 2 things instead of 5?

internal/vmcapabilities/legacysecret.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	f.logger.Debugf(ctx, "found credential secret %s/%s", credentialSecret.Namespace, credentialSecret.Name)
+
+	return credentialSecret, nil
+}
+
+// getLegacyCredentialSecret tries to find a Secret in the default credentials namespace but labeled with the organization name.
+// This is needed while we migrate everything to the org namespace and org credentials are created in the org namespace instead of the default namespace.
+func (f *FactoryImpl) getLegacyCredentialSecret(ctx context.Context, ctrlClient client.Client, objectMeta metav1.ObjectMeta) (*v1alpha1.CredentialSecret, error) {
```
njuettner, line 128: this looks similar to the function above, is that right?

njuettner (approved):

internal/vmcapabilities/legacysecret.go
whites11: this one looks specifically for the k8s secret object, while the other one is the entry point for the whole feature and returns the actual credentials (potentially using this function to get those from a secret).

internal/capzcredentials/capz.go
whites11: good point, refactored it

whites11:
I'll go on with Nick's approval.

PR #373, opened 2022-03-16
Remove validation of `location` field in `AzureMachine` CRs.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/21307


pkg/azuremachine/validate_create.go
```diff
@@ -32,17 +32,12 @@ func (h *WebhookHandler) OnCreateValidate(ctx context.Context, object interface{
...
-		return microerror.Mask(err)
-	}
-
 	vmcaps, err := h.vmcapsFactory.GetClient(ctx, h.ctrlClient, cr.ObjectMeta)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	supportedAZs, err := vmcaps.SupportedAZs(ctx, cr.Spec.Location, cr.Spec.VMSize)
+	supportedAZs, err := vmcaps.SupportedAZs(ctx, h.location, cr.Spec.VMSize)
```
whites11, line 40: In order to check if the availability zones specified are valid, we use the global location field (ultimately is a flag of the admission controller). This is the only location that we accept for the cluster so it should be ok.

fiunchinho: Good solution

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Remove validation of `location` field in `AzureMachine` CRs.
```
fiunchinho, line 12: Maybe worth mentioning the change that uses the location passed as parameter to the admission controller rather than the value in the CR.

whites11: added

fiunchinho (approved):

whites11:
tested, works for Lorenzo's use case

PR #380, opened 2022-03-30
Remove mutations on fields that are already defaulted on 'kubectl-gs'

fiunchinho:
We have some defaulting for labels like the GS `Releases` and the components that will reconcile the CRs. I've moved that [defaulting to the source](https://github.com/giantswarm/kubectl-gs/pull/708) so that when we generate those CRs they already have the right labels.

Notice that this PR is not removing the validation, it's just removing the defaulting. I think that makes our admission controller simpler and less _magical_. The UX should be as good: customers use `kubectl-gs` to generate templates, and if they fiddle with the CRs they will get an error if something that we require is missing.

My hidden agenda here is that I'm removing more parts of the admission controller that are not compatible with dry-run requests, getting closer to make our admission controller work with `flux`.

What do you think?


whites11 (approved):
yeah makes sense

PR #381, opened 2022-03-30
Remove validation of release label on AzureCluster

fiunchinho:
In our `azure-admission-controller` we make sure `AzureCluster` has the same `release` label than `Cluster`, but unless I'm missing something... we don't use the `release` label from `AzureCluster` for anything really on `azure-operator`, we use the one in the `Cluster` CR and then in `MachinePools`/`AzureMachinePools` to see how the upgrade is going. So I think we don't need to validate that label at all.


fiunchinho:
Tested it on `godsmack` by creating a cluster and upgrading it. I didn't see any errors, apart from `context deadline exceeded` coming from the admission controllers every once in a while, but I believe it's not related to these changes.

whites11 (approved):

PR #405, opened 2022-06-10
Resiliency

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/22365

the admission webhook is too slow.
I found out the root cause is VPA setting the CPU limit to low and making everything super slow.

This PR adds a min value to the CPU to ensure a base level of speed.


njuettner (approved):
lgtm

PR #412, opened 2022-06-23
Ignore capi errors

whites11:
When upgrading a v1alpha3 cluster to a v1beta1 one, the newly added fields are being created, but capi admission controller blocks that.
This prevents us from upgrading legacy clusters.

This PR fixes just that.

### DIFF
--- a/pkg/azurecluster/validate_update.go
+++ b/pkg/azurecluster/validate_update.go
@@ -36,6 +36,13 @@ func (h *WebhookHandler) OnUpdateValidate(ctx context.Context, oldObject interfa
 	err = errors.IgnoreCAPIErrorForField("spec.SubscriptionID", err)
 	err = errors.IgnoreCAPIErrorForField("spec.ControlPlaneEndpoint.Host", err)
 	err = errors.IgnoreCAPIErrorForField("spec.ControlPlaneEndpoint.Port", err)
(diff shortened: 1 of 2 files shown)


pkg/azurecluster/validate_update.go
```diff
@@ -36,6 +36,13 @@ func (h *WebhookHandler) OnUpdateValidate(ctx context.Context, oldObject interfa
 	err = errors.IgnoreCAPIErrorForField("spec.SubscriptionID", err)
 	err = errors.IgnoreCAPIErrorForField("spec.ControlPlaneEndpoint.Host", err)
 	err = errors.IgnoreCAPIErrorForField("spec.ControlPlaneEndpoint.Port", err)
+	isCapi, lerr := generic.IsCAPIRelease(azureClusterOldCR)
```
nprokopic, line 39: azure-admission-controller is ignoring all CRs from CAPI releases, so this check should not be required

nprokopic: e.g. https://github.com/giantswarm/azure-admission-controller/blob/master/pkg/validator/http_handler_factory.go#L71-L83

whites11: ok removed

whites11:
Sorry I need to unblock this. Merging

